PR #4, opened 2016-11-02
Make prop types more specific

thomasthesecond:
This PR replaces vague prop types `React.PropTypes.array` and `React.PropTypes.object` with `React.PropTypes.arrayOf`, `React.PropTypes.objectOf` and `React.PropTypes.shape`.

It also updates the eslint config to throw errors if any of the vague prop types are used.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,7 @@
     "es6": true
   },
   "rules": {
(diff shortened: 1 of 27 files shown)


nicrocs:
👍

PR #5, opened 2016-11-02
Remove Rizzo Next

thomasthesecond:
Rizzo Next was only being used for settings. Now the settings are in Backpack. This should make things easier.

Also added a few missing things and updated ListItemBookable to reflect a change made in the POIs app.


nicrocs:
👍

PR #8, opened 2016-11-03
Update icons

thomasthesecond:
This PR takes all of the current icons out of Rizzo Next and puts them into a React component. The SVG image map has been replaced and should make it easier to maintain and use the icons.

The old way to use the icons was:

`<Icon name="bookmark" />`

The new way to use the icons is:

`<Icon.Bookmark />`

Here are the props that can be passed along with the icon:

- `viewBox`: Most of the interface icons have a viewbox of "0 0 32 32" and any new icons should try to fit this size, but if not, then it can be changed with this prop
- `className`: If a styling hook is needed for an external stylesheet
- `fill`: Change the color of the icon; by default, the fill uses the CSS variable `currentColor`
- `width`: Change the width of the icon
- `height`: Change the height of the icon
- `label`: Provide descriptive text for the icon (aria-label)
- `style`: Apply additional CSS styles to the icon


jcreamer898:
Cool!

src/components/icon/index.jsx
```diff
@@ -1,191 +1,697 @@
...
 Icon.defaultProps = {
-  name: "",
+  viewBox: "0 0 32 32",
+};
 
-  size: "",
+// Interface
 
-  inline: "",
+exports.ArrowLeft = props => (
```
nicrocs: do we want to do exports in this file? Would it work better in it's own file?

thomasthesecond: I thought about making a file for each icon, but that'd be a lot of files, so I didn't. I also thought it would be easier having all the SVGs in a single file.

nicrocs: Makes sense, let's go with this for now.

PR #9, opened 2016-11-03
Copy settings to dist, bump version

thomasthesecond:
settings.json file was not being copied from src to dist which caused an error when importing the components.


nicrocs:
👍

PR #12, opened 2016-11-09
Update component styles

thomasthesecond:
Without a global stylesheet, some components fail to render properly.

Some of the updates include:
- Replacing rem with px because each app may define its own global font
size, which will render the components inconsistently
- Adding reset styles to certain elements, such as removing link
underlines and removing padding, margin on unordered lists
- Setting a font-family on components because there is no global
stylesheet to inherit from
- Update prop types validation

### DIFF
--- a/src/components/contentHeader/index.jsx
+++ b/src/components/contentHeader/index.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import radium from "radium";
+import assign from "object-assign";
 import settings from "../../../settings.json";
 import Heading from "../heading";
 import MoreLink from "../moreLink";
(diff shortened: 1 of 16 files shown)


src/components/moreLink/index.jsx
```diff
@@ -50,6 +49,7 @@ const styles = {
  */
 function MoreLink({ href, size, children, onClick, caps, style }) {
   const Element = href ? "a" : "button";
+  const iconStyle = assign({}, styles.icon.base, size ? styles.icon.size[size] : {});
```
nickfranciosi: If you are returning an empty object if there is no size you can do 
`size && styles.icon.size[size]`

src/components/contentHeader/index.jsx
```diff
@@ -52,12 +53,14 @@ function ContentHeader({ title, heading, border, moreLink }) {
     "ContentHeader clearfix" :
     "ContentHeader";
 
+  const headingStyle = assign({}, hasLink ? styles.heading.base : {}, heading.override);
```
nickfranciosi: same here with the `&&`

src/components/moreLink/index.jsx
thomasthesecond: I think I've had issues where that syntax returns falsey which wouldn't always give me the right type, but you're right I think here it will be ok.

PR #24, opened 2016-11-23
Add container component

thomasthesecond:
Experiments with styled-components


thomasthesecond:
Cool man, thanks. I like it, too.

PR #34, opened 2016-12-02
update props / update radium

nickfranciosi:
bump up version of rizzo and fix the prop type warning once and for all 🎉


thomasthesecond:
Cool. After this is merged, I'll do a 1.0.0 release for backpack

PR #41, opened 2016-12-08
Add Number Marker Xtra Small size

nickfranciosi:


src/components/numberMarker/index.jsx
```diff
@@ -21,6 +21,13 @@ const styles = {
     },
 
     size: {
+      xSmall: {
+        height: "2.4rem",
+        width: "2.4rem",
+        lineHeight: "2.4rem",
```
thomasthesecond: Can we use px for these values? I've been slowly replacing rem with px because our components aren't gonna look right in apps that don't have `html { font-size: 10px }`

nickfranciosi:
@thomasthesecond Updated with those changes

PR #43, opened 2016-12-12
fix article list item width

nickfranciosi:
not exactly sure what is happening here, but taking it out fixes the issue on ie11. 

Feel free to make suggestions on the best way to handle this.


thomasthesecond:
I think this will break the center aligned content, but I'll take a look.

thomasthesecond:
The anchor was `display: flex` to keep the "more link" pinned to the bottom of each callout regardless if one callout had more text than the others. Since the "more link" was removed from the design, I removed it from this component and set the anchor's display to `block` since it's wrapping.

And just an FYI, the IE 11 flexbox bug was fixed by putting `width: 100%` on the `h3` and `p` elements.

PR #46, opened 2016-12-16
fix link for sights list item

nickfranciosi:
took out relative link path

### DIFF
--- a/src/components/sightsListItem/index.jsx
+++ b/src/components/sightsListItem/index.jsx
@@ -71,7 +71,7 @@ function SightsListItem({
   markerNumber,
 }) {
   return (
-    <a style={styles.linkContainer} href={`/${slug}`}>
+    <a style={styles.linkContainer} href={slug}>
       <div style={styles.imageContainer}>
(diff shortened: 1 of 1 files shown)


src/components/sightsListItem/index.jsx
```diff
@@ -71,7 +71,7 @@ function SightsListItem({
   markerNumber,
 }) {
   return (
-    <a style={styles.linkContainer} href={`/${slug}`}>
+    <a style={styles.linkContainer} href={slug}>
```
thomasthesecond: Hmm I could see this being an issue if an absolute URL was ever passed in. Can we just pass in a relative URL instead of expecting one?

nickfranciosi: @thomasthesecond the component can still take a relative link, you would just put in the beginning slash instead of it being default. This way you can do both

thomasthesecond: Oh man, I'm sorry, I totally missed that you removed the slash, not added it.

nickfranciosi: 👍

PR #67, opened 2017-01-24
Update social sharing components

thomasthesecond:
- Added Reddit and Facebook Messenger as sharing networks
- Created a container component to handle the logic for opening pop-up windows for Facebook, Reddit and Twitter


src/components/socialShare/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+};
+
+const SocialShare = ({ headingText, href, onClick }) => (
+  <div className="SocialShare">
+    <Heading size="tiny" weight="thick" level={5} caps>
+      {headingText}:
+    </Heading>
+
+    <div style={styles.container}>
+      <SocialIconButton
```
nickfranciosi: @thomasthesecond just wonder if we should be passing in an array of config objects and mapping over them instead of coding all the buttons

src/components/shareMenu/index.jsx
```diff
@@ -186,21 +111,21 @@ class ShareMenu extends React.Component {
           <Flyout arrow={flyoutArrow} fill>
             <ShareMenuItem
               network="twitter"
-              href={twitter}
+              href={href.twitter}
```
nickfranciosi: @thomasthesecond just wonder if we should be passing in an array of config objects and mapping over them instead of coding all the items (also see below)

thomasthesecond:
Made some changes:

A `networks` object gets passed as a prop to `ShareMenu` and `SocialShare` from the container component `SocialShareContainer`. This allows us to loop over the networks to render each button. Also, I added a `hide` prop which accepts an array of networks to keep hidden from the list; this was something I took from our social buttons component in Rizzo.

PR #72, opened 2017-01-30
base card / video card/ price card

nickfranciosi:
Took the card from homepage / interest page and tried to make a more generic base for us to create different cards. 

Created some helper functions/ components that were needed for the video card specifically for when nested links are needed. 

Kept the legacy card for reference or if we need to pull it into homepage/ interest pages in the short term.


thomasthesecond:
Closing this since we have the card components built with Radium for now.

PR #73, opened 2017-01-30
Delete styled-components dependency

thomasthesecond:


jcreamer898:
Wow, that happened faster than I could have ever imagine.

mxstbr:
`styled-components` maintainer here—I found this by chance, out of interest why did you revert the integration of `styled-components` and went back to inline styles?

Would love to hear your feedback on the library, and where it came short of your expectations! (sorry if I missed a relevant issue)

thomasthesecond:
@mxstbr Thanks for reaching out! I loved using styled components and personally, prefer them much more than inline styles, but we made the decision to remove styled-components because server-side rendering is a big thing for us. We've been following the discussion https://github.com/styled-components/styled-components/issues/124 and I think our plan is to re-visit when that's an option (at least I would like us to).

Perhaps @jcreamer898 or @nickfranciosi can weigh in with their thoughts as well.

mxstbr:
That's great to hear!

I'd hate for that to be the last call, just because we haven't shipped version two with SSR support yet. 😢
If you're willing to give it a try, there's a v2 pre-release, plus an awesome new [Babel plugin](https://github.com/styled-components/babel-plugin-styled-components) that gives you a much better DX by putting the component name into the className.

You can get the pre-release like this:

```
npm install --save --save-exact styled-components@next
```

And can use the API that's been discussed in that issue for SSR:

```JS
import { styleSheet } from 'styled-components'

styleSheet.getCSS()
styleSheet.reset()
```

(we _might_ change the API in the future, so make sure to lock your version!)

thomasthesecond:
That's awesome! Thanks so much for sharing, I'm looking forward to checking out the v2 pre-release.

mxstbr:
Oh I almost forgot, here you can see an example of this that works in case you get stuck somewhere: https://github.com/styled-components/styled-components/blob/v2/example/ssr.js

jcreamer898:
@mxstbr we'll definitely give it a try, do you have any idea about how stable the `next` branch is? We'll be shipping this stuff into production pretty soon. :)

mxstbr:
> do you have any idea about how stable the next branch is? 

Stable enough for sure, I use it myself! 😉 we're planning to release v2 soon, so there's only some small new APIs to be added, mostly.

Worst case scenario you ping me (contact@mygithubhandle.com) and I push a new bugfix release!

PR #79, opened 2017-01-31
list items w/ thumbnail

nickfranciosi:
adds list item component with thumbnail that is needed for lp.tv

### DIFF
--- a/src/utils/icon.jsx
+++ b/src/utils/icon.jsx
@@ -1,0 +1,8 @@
+import React from "react";
+import Icon from "../components/icon";
+
+export const iconFromString = (iconName, props) => React.createElement(Icon[iconName], {
+  ariaHidden: true,
+  className: "Icon",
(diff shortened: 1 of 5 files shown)


src/utils/icon.jsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import Icon from "../components/icon";
+
+export const iconFromString = (iconName, props) => React.createElement(Icon[iconName], {
+  ariaHidden: true,
+  className: "Icon",
+  ...props,
+});
```
thomasthesecond: You have this in the other PR, I don't think it will hurt anything though? Just not sure if it was intentional to have it twice.

nickfranciosi: definitely wasn't. since it was a new folder I must have missed this when I switched branches.
I'll take care of it. Thanks!

PR #87, opened 2017-02-02
Update checkbox component

thomasthesecond:
- Moved to root components folder
- Updated `size` prop accept numbers to size checkbox
- Tightened up styles
- Added `rounded` prop


src/components/checkbox/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+            style={[
+              styles.input,
+              {
+                height: `${size}px`,
+                width: `${size}px`,
+              },
+            ]}
+            value={value}
+            name={_.kebabCase(name)}
+            onClick={(event) => this.onClick(event, value, _.kebabCase(name))}
```
nickfranciosi: Can we grab ```name``` and ```value``` from inside the function call itself. Then we can just put ```onClick={this.onClick}``` which won't create a new function on re-render. Then inside the ```this.onClick``` function you can pull name / value off of state or props

PR #98, opened 2017-02-08
Added SpotlightZone and SectionHeader components

harterc1:
screen shot 2017-02-08 at 2 26 53 pm
screen shot 2017-02-08 at 2 27 07 pm
screen shot 2017-02-08 at 2 45 59 pm
screen shot 2017-02-08 at 2 46 07 pm


thomasthesecond:
Initially, I noticed some things that our linter should have picked up. This is my fault because I don't have a pre-commit hook to check linting. I'll make sure to add that, but in the meantime, I'll write inline comments where I know things will cause an error.

I'm not sure what editor you use, but for Atom, there is a linter package for ESLint which is handy because it notifies you of warning and errors inline. We also use editorconfig to help with formatting, so I would recommend installing the plugin for your editor http://editorconfig.org/#download

As I write this, I realize this would be good information to know beforehand, so I will put it in the README.

src/components/sectionHeader/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+const styles = {
+  container: {
+    textAlign: "center"
+  },
+
+  heading: {
+    marginBottom: "1.3rem",
+    lineHeight: 1.3,
+    fontSize: "2.8rem",
```
thomasthesecond: Can we use `px` instead of rem for all instances?

Now that we have our core.css file in backpack, it's not as much of an issue, but the reasoning for `px` is that if these components are used inside of some other app that doesn't have `html { font-size: 10px }`, then the component wouldn't look right. I know we currently have a mix of `px` and `rem`, but the goal is to use `px`.

src/components/sectionHeader/index.jsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import radium from "radium";
+import Heading from "../heading";
+import settings from "../../../settings.json";
```
thomasthesecond: This isn't required, more of an FYI, but you can just import the settings you need like so: `import { color } from "../../../settings.json;`

src/components/sectionHeader/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+const SectionHeader = ({ title, heading }) => {
+  heading = heading || {};
+  heading.size = heading.size || "large";
+  heading.weight = heading.weight || "extraThin";
+
+  return (
+    <div style={styles.container}>
+      <Heading {...heading} override={styles.heading}>
+        {title}
```
thomasthesecond: Can we use React's `children` prop here instead of `title`? So you would have this when you used the component:

```
<SectionHeader>
  Your title here
</SectionHeader>
```

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+      paddingLeft: gutter("static"),
+      paddingRight: gutter("static")
+    },
+
+    [`@media (min-width: ${settings.media.min["1080"]})`]: {
+      paddingLeft: gutter("static", 1, 2),
+      paddingRight: gutter("static", 1, 2)
+    }
+  }
+}
```
thomasthesecond: Here's another place where the linter would pick up the missing semi-colon and then trailing commas on lines 28, 29, 30. My fault, I'll get that pre-commit hook in soon.

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+
+const styles = {
+  container: {
+    fontFamily: font("benton"),
+    position: 'relative',
```
thomasthesecond: We use double quotes instead of single. This occurs in other places.

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+import radium from "radium";
+import { Link } from "react-router";
+import settings from "../../../settings.json";
+import { gutter } from "../../utils/grid";
+import font from "../../utils/font";
+import { rgb } from "../../utils/color";
+import Heading from "../heading";
+import { Play } from "../icon";
+
+const spotlightTitleGray = "#e4e4e4";
```
thomasthesecond: Use `settings.color.gray` here instead of `spotlightTitleGray`.

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+}
+
+const SpotlightZone = ({
+  zone, 
+  category, 
+  title, 
+  paragraph, 
+  href, 
+  imageUrl, 
+  backgroundImageUrl}) => {
```
thomasthesecond: Can we break the `}) => {` to a new line?

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+  
+  let backgroundStyle = styles.background;
```
thomasthesecond: Prefer `const` instead of `let` when the variable isn't re-assigned.

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+  backgroundStyle.backgroundImage = `url('${backgroundImageUrl}')`;
+
+  return (
+    <div className="SpotlightZone" style={styles.container}>
+      <style dangerouslySetInnerHTML={markup(css)} />
+      <div style={backgroundStyle}></div>
+      <div style={styles.backgroundOverlay}></div>
```
thomasthesecond: In React, when you have an empty element like on lines 174 and 175, you can write it as a self-closing element like so:

`<div style={backgroundStyle} />`

and I think we have a lint rule for this as well.

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+        <div style={styles.rightContent}>
+          <Link to={href} style={styles.link}>
+            <img style={styles.image} src={imageUrl} />
+            <div className="SpotlightZone-imageOverlay" style={styles.imageOverlay}>
+              <Play style={styles.playIcon} />
+            </div>
+          </Link>
+        </div>
+      </div>
+    </div>
```
thomasthesecond: In this block of code, can we add some whitespace and break things to newlines to help with readability. For example this

`<div style={styles.zone}>{zone}</div>`

can become this:

```
<div style={styles.zone}>
  {zone}
</div>
```

src/components/spotlightZone/index.jsx
```diff
@@ -0,0 +1,220 @@
...
+  title: "",
+  paragraph: "",
+  href: "",
+  imageUrl: "",
+  backgroundImageUrl: "",
+};
+
+SpotlightZone.styles = styles;
+
+export default radium(SpotlightZone);
```
thomasthesecond: Can we add a newline to the end of the file here? This will get rid of github's note about it. I noticed it on the sectionHeader/index.jsx file above too.

stories/index.jsx
```diff
@@ -1201,6 +1212,14 @@ storiesOf("Sectional nav", module)
     </StyleRoot>
   ));
 
+storiesOf("Section Header", module)
+  .addDecorator(withKnobs)
+  .add("Default", () => (
+    <StyleRoot>
+      <SectionHeader title={text("title", "Top experiences in Vietnam")}/>
```
thomasthesecond: As i mentioned about about using `children`, this would then become
`<SectionHeader>{text("Title", "Top experiences in Vietnam")}</SectionHeader>`

src/components/spotlightZone/index.jsx
harterc1: So trailing commas are a good thing?  (in the linters eyes)

thomasthesecond: Yep, they're removed when Babel compiles the code.

https://github.com/airbnb/javascript#commas--dangling

PR #100, opened 2017-02-09
Add photo gallery from POIs, articles

thomasthesecond:
Will be used in the travel news app.

I don't love how I'm pulling in the styles, but it was the quickest way to get them in. Any thoughts?

### DIFF
--- a/src/components/photoGallery/styles.js
+++ b/src/components/photoGallery/styles.js
@@ -1,0 +1,582 @@
+export default `
+/*! PhotoSwipe main CSS by Dmitry Semenov | photoswipe.com | MIT license */
+.pswp {
+    display: none;
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    left: 0;
+    top: 0;
+    overflow: hidden;
+    -ms-touch-action: none;
+    touch-action: none;
+    z-index: 1500;
+    -webkit-text-size-adjust: 100%;
+    -webkit-backface-visibility: hidden;
+    outline: none
+}
+
+.pswp img {
+    max-width: none
+}
+
+.pswp--animate_opacity {
+    opacity: .001;
+    will-change: opacity;
+    transition: opacity 333ms cubic-bezier(.4,0,.22,1)
+}
+
+.pswp--open {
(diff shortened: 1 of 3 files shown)


src/components/photoGallery/styles.js
```diff
@@ -0,0 +1,582 @@
+export default `
+/*! PhotoSwipe main CSS by Dmitry Semenov | photoswipe.com | MIT license */
```
nickfranciosi: @thomasthesecond you could install photoswipe via npm and then import the css from the dist folder into your component. I tried this method with react-slick today and it seemed to work well. not sure if it will cause any other issues though.

jcreamer898: Agreed.

thomasthesecond:
I thought about that, but then how does that work within backpack? I need to bundle the styles with the component.

ImageCarousel uses react-slick, but I see no stylesheet https://github.com/lonelyplanet/backpack-ui/blob/master/src/components/imageCarousel/index.jsx

@jcreamer898 are you talking about doing this on a per-app basis? Backpack doesn't use webpack. I think the styles need to be included with the component somehow.

jcreamer898:
I see what you mean, good point. Hmm...

jcreamer898:
It may just be that for this particular component, the consuming app has to import the styles and the react component.

nickfranciosi:
@thomasthesecond @jcreamer898  https://github.com/lonelyplanet/backpack-ui/blob/nf-masthead/src/components/mastheadSlider/index.jsx#L4

This PR isn't ready but I imported the slick carousel css directly into the masthead slider and it seems to be working. Let me know if you think this will break later.

nickfranciosi:
so I tried to pull that component into dotcom-video and I am getting an error with css import so i guess that wont work.... 
It would be ideal to find a solution within backpack so that we don't have all these extra dependencies that we need to remember to pull into the header. I will dig around a little bit

thomasthesecond:
Need to `import "react-photoswipe/lib/photoswipe.css";` into common.js

PR #101, opened 2017-02-09
Add lint script

thomasthesecond:
I'll be adding the pre-commit in another pull request. It's just that there are a few lint errors that need fixed up before the pre-commit hook can be pushed to master. I wanted to get the lint script in there so at least it can be done manually for now.


jcreamer898:
Clean code?! Pfffsh.

PR #102, opened 2017-02-09
default masthead / masthead slider / slide

nickfranciosi:
Adds default styles and hooks to customize masthead slider
Adds default Slide component, not sure if we want to customize this very much or just create different slide types for different needs to pass into the masthead since it takes any slide component


thomasthesecond:
Can we move mastheadSlider/styles.js into mastheadSlider/index.jsx

stories/index.jsx
```diff
@@ -1,6 +1,8 @@
 import React from "react";
 import { StyleRoot } from "radium";
 import "leaflet/dist/leaflet.css";
+import "slick-carousel/slick/slick.css";
+import "slick-carousel/slick/slick-theme.css";
```
thomasthesecond: I'm still not sold on this being the best approach for including third-party component styles. I'm not sure what the answer is.

nickfranciosi: yeah, right now continuing on the path of including the css in the separate projects

stories/index.jsx
```diff
@@ -817,6 +822,64 @@ storiesOf("Map marker", module)
...
+  ))
+  .add("With Slider", () => {
+    const items = [{
+      adPosition: "",
+      image: "https://lonelyplanetstatic.imgix.net/copilot%2Fimages%2FR2V0dHlJbWFnZXMtNjA3Njk2MTAzX2Z1bGwuanBnV2VkIEZlYiAwMSAyMDE3IDA5OjAxOjU1IEdNVCswMDAwIChVVEMp.jpg?q=40&sharp=10&w=2500",
+      type: "Featured article",
+      headline: "Honeymoon hacks: a guide for newly-weds abroad",
+      description: ["Item 1", "Item 2"],
+      callToAction: {
+        text: "HAPPILY EVER AFTER",
```
thomasthesecond: Are we doing `text-transform: uppercase` on every instance where you have all caps? If not, can we and then remove the caps? I believe our editorial guidelines say to capitalize the first letter only.

nickfranciosi: I just copied that object from something, They shouldn't be uppercase. just test text

src/components/masthead/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+    zIndex: zIndex.slideshowSlide,
+    [`@media (min-width: ${media.min["720"]})`]: {
+      height: "100vh",
+    },
+  },
+
+  isUnderGlobalHeader: {
+    marginTop: "-5rem",
+    [`@media (min-width: ${media.min["720"]})`]: {
+      marginTop: "-13rem",
```
thomasthesecond: px

nickfranciosi:
@thomasthesecond for mastheadSlider/styles, that is really long but if that is where we want it regardless of length then that is cool with me.

src/components/masthead/index.jsx
nickfranciosi: I am having to match exactly what is in rizzo-next which uses rems so I don't know if it would be too easy to get off with values. What do you think?

src/components/mastheadSlider/styles.js
```diff
@@ -0,0 +1,72 @@
...
+export const rules = {
+  ".slick-dots": {
+    marginRight: "auto",
+    marginLeft: "auto",
+    height: "34px",
+    position: "absolute",
+    textAlign: "center",
+    right: 0,
+    left: 0,
+    bottom: 32,
```
thomasthesecond: px

src/components/mastheadSlider/styles.js
```diff
@@ -0,0 +1,72 @@
...
+    maxWidth: "10rem",
+  },
+  ".slick-dots button:before": {
+    opacity: 1,
+    fontSize: "1rem",
+    color: "rgba(255,255,255, 0.37)",
+  },
+  ".slick-dots .slick-active button:before": {
+    opacity: 1,
+    color: "rgba(255,255,255, 1)",
```
thomasthesecond: Can we use our rgba function and color.white here and line 37?

src/components/mastheadSlider/styles.js
```diff
@@ -0,0 +1,72 @@
...
+    marginLeft: "auto",
+    height: "34px",
+    position: "absolute",
+    textAlign: "center",
+    right: 0,
+    left: 0,
+    bottom: 32,
+  },
+  ".slick-dots li": {
+    width: "1.2rem",
```
thomasthesecond: px instead of rem in all instances in this file

src/components/slide/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+        >
+          <a href={callToAction.link} style={styles.link} >
+            {headline}
+          </a>
+        </Heading>
+
+        {adPosition &&
+          <div
+            id={`video-home-sponsor-advert-${adPosition}`}
+            sstyle={styles.adContainer}
```
thomasthesecond: extra "s"

src/components/masthead/index.jsx
thomasthesecond: Ok, that's fine then. can we put a comment about it?

nickfranciosi:
@thomasthesecond should be good now

PR #103, opened 2017-02-09
Create shelf, swiper and grid components for cards and tiles

thomasthesecond:
I might need to re-write the swiper to use react-slick. I need to see if provides me the same functionality.


src/components/cardShelfVideoSwiper/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+    width: "calc(100% + 100px)",
+    padding: "30px 50px 80px",
+    marginTop: "-30px",
+    marginLeft: "-50px",
+    overflow: "hidden",
+    position: "relative",
+  },
+};
+
+const swiperOptions = {
```
nickfranciosi: @thomasthesecond  does this library have responsive options. Looking at the comp it appears the arrows disappear.

src/components/cardShelf/index.jsx
```diff
@@ -0,0 +1,5 @@
+export CardShelf from "./cardShelf";
+export CardShelfContent from "./cardShelfContent";
+export CardShelfHeader from "./cardShelfHeader";
+
+export default from "./cardShelf";
```
nickfranciosi: Not sure CardShelf needs to be exported as both a named export and default

src/components/cardShelfVideo/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+);
+
+CardShelfVideo.propTypes = {
+  children: (props, propName, componentName) => {
+    const prop = props[propName];
+    let error = null;
+
+    React.Children.forEach(prop, (child) => {
+      if (child.type !== CardVideo) {
+        error = new Error(`${componentName} children should be of type "CardVideo".`);
```
nickfranciosi: Will that taller card we need to build also be of type CardVideo? We may need to loosen this up a bit

src/components/cardShelfVideoSwiper/index.jsx
thomasthesecond: It does

src/components/cardShelf/index.jsx
thomasthesecond: Will remove

PR #108, opened 2017-02-13
add segment component

nickfranciosi:
adds segment which from copilot that handles vertical spacing and border between sections

### DIFF
--- a/src/components/segment/index.jsx
+++ b/src/components/segment/index.jsx
@@ -1,0 +1,87 @@
+import React, { PropTypes } from "react";
+import radium from "radium";
+import { color, media } from "../../../settings.json";
+import { rgb } from "../../utils/color";
+
+const styles = {
+  container: {
+    default: {
+      marginBottom: "88px",
+      paddingBottom: "88px",
+      [`@media (min-width: ${media.min["720"]})`]: {
+        marginBottom: "128px",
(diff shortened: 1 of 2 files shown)


src/components/segment/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+  </div>
+);
+
+Segment.propTypes = {
+  id: PropTypes.string,
+  border: PropTypes.bool,
+  halfSpace: PropTypes.bool,
+  showBorderAtBreakpoint: PropTypes.number,
+  showAbove: PropTypes.bool,
+  style: PropTypes.object,
```
thomasthesecond: Need an objectOf here

src/components/segment/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+}) => (
+  <div
+    className="Segmented"
+    id={id}
+    style={[
+      styles.container.default,
+      border && styles.container.hasBorder,
+      halfSpace && styles.container.halfSpace,
+      showBorderAtBreakpoint && breakpointBorder(showBorderAtBreakpoint, showAbove),
+      style && style,
```
thomasthesecond: Can just be `style`

nickfranciosi:
@thomasthesecond should be good now.

PR #110, opened 2017-02-13
Added VideoEmbed component

harterc1:


src/components/videoEmbed/index.jsx
```diff
@@ -0,0 +1,185 @@
+import React from "react";
+import radium from "radium";
+import _ from "lodash";
```
jcreamer898: Instead of importing all of lodash, you can import the specific methods you need....

```js
import get from "lodash/get";
import kebabCase from "lodash/kebabCase";

const _ = { get, kebabCase };
```

That saves size in the webpack bundle.

jcreamer898:
One small cleanup. Nice work!

PR #114, opened 2017-02-13
scope video scroller css

nickfranciosi:


src/components/cardShelfVideoSwiper/index.jsx
```diff
@@ -15,7 +15,7 @@ import PaginatorButton from "../paginatorButton";
 const styles = {
   slider: {
     width: "calc(100% + 100px)",
-    padding: "30px 50px 80px",
+    padding: "30px 50px 0",
```
thomasthesecond: Why was bottom padding removed? This was here to allow the box-shadow to be visible on hover. Otherwise, overflow: hidden clips it

nickfranciosi: I forgot about the version with the shadow. the padding is messing up the vertical spacing on the page. The segment component takes care of how things should be space out, but this extra padding makes it look awkward. I had to some negative margin with the tallCarousel component to get this to maintain the proper distance

thomasthesecond: You could do this:

```
<div style={{ marginBottom: "-80px" }}>
  <CardShelfVideoSwiper />
</div>

<div>something else</div>
```

thomasthesecond: or actually maybe just add it to the component

https://github.com/lonelyplanet/backpack-ui/blob/0be442b3176c154481c95bb6bcc0bc464f56e64f/src/components/cardShelfVideoSwiper/index.jsx#L17-L22

nickfranciosi:
@thomasthesecond I removed it

PR #115, opened 2017-02-13
pull in spacing from copilot section header

nickfranciosi:


src/components/sectionHeader/index.jsx
```diff
@@ -6,6 +6,10 @@ import { color, media } from "../../../settings.json";
 const styles = {
   container: {
     textAlign: "center",
+    marginBottom: "40px",
+    [`@media (min-width: ${media.min["720"]})`]: {
+      marginBottom: "64px",
+    },
```
thomasthesecond: I think this should be passed in via the style prop. I don't think our components should have any margins included.

nickfranciosi: The consistency for spacing on these should always be the same. If we do it that way we will have to pass it in every single time we use it on the page

PR #127, opened 2017-02-15
Added autoPlay prop to VideoEmbed

harterc1:


src/components/videoEmbed/index.jsx
```diff
@@ -175,9 +181,14 @@ class VideoEmbed extends Component {
...
   id: PropTypes.string.isRequired,
   videoId: PropTypes.string.isRequired,
+  autoPlay: PropTypes.bool,
   override: PropTypes.oneOfType([
     PropTypes.object,
   ]),
 };
 
+VideoEmbed.defaultProps = {
+  autoPlay: true,
```
thomasthesecond: Generally, on boolean props, the default is set to false so that you may write
`<VideoEmbed autoplay />` rather than `<VideoEmbed autoplay={false} />`

Even though we want our videos to autoplay, I still think autoplay should be "opt-in", mimicking HTML5's `video` https://msdn.microsoft.com/en-us/library/ff974744(v=vs.85).aspx

and also I wouldn't use camelCase in `autoplay` in order to match the HTML5 video attribute which is all lower case.

harterc1: good points.  will update.

PR #130, opened 2017-02-16
Add onClick prop to listItem

thomasthesecond:


thomasthesecond:
It goes to the URL and triggers the onClick

https://github.com/lonelyplanet/backpack-ui/blob/master/src/components/typeSelector/index.jsx#L220-L229

In the case I'm using it, I'm bypassing the `to` with `window.location` because it won't link to the POIs on the staging URL. https://github.com/lonelyplanet/dotcom-video/pull/70

thomasthesecond:
If there's a better way of using it, then let me know. I think this is a downside of react-router's `Link` when we aren't using it on production

PR #131, opened 2017-02-16
Trying to fix the masthead

jcreamer898:
Plz help, it's so dark.


nickfranciosi:
@thomasthesecond  this fixes the issue, not sure if you wanted to take a look before we merged it in

PR #137, opened 2017-02-21
add validation component

nickfranciosi:
# Validation Component

Wraps any number of inputs and validates them on either a set of build in rules or custom passed in rules. The child of the Validation component must be a function and that function is passed several arguments to work with. 

### Props Passed Through
- `validations`
This is a required object that maps input fields to what rules their values need to adhere to. The keys need to match the `name` field on the inputs. The value needs to be an array of strings that correspond to the rules that need to be used for that input. 
```javascript
     const validations = {
        firstName: ["required"],
        email: ["required", "email"],
       password: ["required", "min:3", "max: 15"]
   }
```
- `rules`
Rules is an optional object that can be passed through to add custom validations. These custom validations are then referenced in the `validations` prop. A rule consists of two things a `test` function and a `message` function. The key of the object is the name to reference the rule by. 
The `test` function will be passed the value of the input at the point you validating and should evalute to true if the value is valid and false if it is not, and the `message` function will receive the name of the field to in order to display in any error message you may want to create. To create your own rule : 
```javascript
     const rules = {
        coolRule: {
          test: (val) => val.indexOf("cool") >= 0,
          message: (field) => `${field} should contain the word cool`,
        },
     }; 
    const validations = {
        firstName: ["required", "coolRule"],
    };
```

You can also overwrite just a `test` or `message` of any build in rule by just passing in the item you need to replace
```javascript
     const rules = {
        required: {
          message: (field) => `${field} should be required and this message is custom`,
        },
     }; 
    const validations = {
        firstName: ["required"],
    };
```

### Arguments Passed Through To Function
- `validate`
a function that can be attached to any input event in order to validate that input. The event is passed through and the Validation component validates based on the rules passed in
- `errorMessages`
an object of errorMessages. Keys are based on the inputs `name` field. An input with `name="first"` an access error messages from `errorMessages.first`
- `allValid`
a boolean telling you if all fields attached to the validator are valid
- `errorCount`
total error count
```javascript
  <Validate
      validations={{
        firstName: ["required", "min:3"],
      }}
    >
      {(validate, errorMessages, allValid, errorCount) => (
         <input type="text" onChange={validate} name="firstName" />
         {errorMessages.firstName && errorMessages.firstName.map(val => <p>{val}</p>)}
      )}
 </Validate>
```
## 
# Built In Rules
- `required`
- `email`
- `min`
- `max`

### Rules with arguments
Some built in rules and any custom rules have the ability to take arguments. For example the built in rule for `min:3` Passes the argument `3` into both the `test` function and `message` function.
Rules with arguments are :

- `min`
- `max`

Creating your own rules that take arguments will need to define both the `test` function and `message` function a bit differently

```javascript
       min: {
          test: (arg) => (val) => val.length >= arg,
          message: (arg) => (name) => `${name} must be at least ${arg} characters.`,
      },
```

You have to curry your functions to take the argument first and then return a function that takes the value.

### DIFF
--- a/src/components/validate/index.jsx
+++ b/src/components/validate/index.jsx
@@ -11,6 +11,14 @@ const validationRules = {
     test: (val) => emailRegEx.test(val),
     message: (name) => `${name} must be a valid email.`,
   },
+  min: {
+    test: (arg) => (val) => val.length >= arg,
+    message: (arg) => (name) => `${name} must be at least ${arg} characters.`,
+  },
+  max: {
+    test: (arg) => (val) => val.length <= arg,
+    message: (arg) => (name) => `${name} must ${arg} less than characters.`,
+  },
 };
 
 class Validate extends Component {
@@ -38,12 +46,24 @@ class Validate extends Component {
     });
   }
 
+
   testForValidation(field, value) {
     const fieldRules = this.props.validations[field];
+
     const combinedValidationRules = Object.assign({}, validationRules, this.props.rules);
     return fieldRules && fieldRules.map(rule => {
-      return !combinedValidationRules[rule].test(value) &&
-      combinedValidationRules[rule].message(field);
+      if (rule.indexOf(":") >= 0) {
+        const [funcName, arg] = rule.split(":");
+        if (combinedValidationRules[funcName]) {
+          return !combinedValidationRules[funcName].test(arg)(value) &&
+          combinedValidationRules[funcName].message(arg)(value);
+        }
+      }
+      return (
+      combinedValidationRules[rule] &&
+      !combinedValidationRules[rule].test(value) &&
+      combinedValidationRules[rule].message(field)
+      );
     }).filter(val => val);
   }
 
--- a/src/components/validate/index.jsx
+++ b/src/components/validate/index.jsx
@@ -1,0 +1,64 @@
+import React, { Component, PropTypes } from "react";
+
+const emailRegEx = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+
+const validationRules = {
+  required: {
+    test: (val) => val.length > 0,
+    message: (name) => `${name} is required.`,
+  },
+  email: {
+    test: (val) => emailRegEx.test(val),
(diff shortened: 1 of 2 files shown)


nickfranciosi:
@jcreamer898 talked about not leaving this in backpack but I wanted to make sure that we liked this approach before I pulled it into its own package. Let me know what your guys thoughts on that are.

thomasthesecond:
Very thorough comment! I'll take a look.

src/components/validate/index.jsx
```diff
@@ -0,0 +1,115 @@
+import React, { Component, PropTypes } from "react";
+import merge from "lodash/merge";
+
+const _ = { merge };
+
+const emailRegEx = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
thomasthesecond: Would this be useful as a util?

```
export default function emailRegEx() {
  return /.../;
}
```

thomasthesecond: Well sorry, I guess if this component is not in Backpack, maybe not

src/components/validate/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+const validationRules = {
+  required: {
+    test: (val) => val.length > 0,
+    message: (name) => `${name} is required.`,
+  },
+  email: {
+    test: (val) => emailRegEx.test(val),
+    message: (name) => `${name} must be a valid email.`,
```
thomasthesecond: Would it make sense to name `name` as `email`?

and same for the others? Name the variable the same as the key?

thomasthesecond:
This is cool, I'm all for it. Where do we think it could live? Would it be its own npm module?

jcreamer898:
`react-validate-form` is available. :)

src/components/validate/index.jsx
nickfranciosi: So name is currently the name of the field, but I could also see where the value could be useful to the error message. 

@thomasthesecond  Do you think I should add that as a second pass through value?

nickfranciosi:
@jcreamer898 @thomasthesecond so we are thinking this should be its own module? I am cool with that. `react-validate-form` under lonely planet works for me

jcreamer898:
That's what I'm thinking.

jcreamer898:
What are your thoughts @thomasthesecond?

nickfranciosi:
moved into own package

PR #138, opened 2017-02-21
Read in attributes off children

jcreamer898:


jcreamer898:
How does it remove that ability? You can still do the `rules` prop and overwrite them in exactly the same way as before no?

nickfranciosi:
but I can't create a custom rule called `myRule` and attach it to an input

jcreamer898:
Oh, sure, I think we should support both the `validations` and the `HTML5` attributes.

PR #148, opened 2017-02-28
Add hover scale effect to CoverPhoto

thomasthesecond:
In card, tile and list item components, the CoverPhoto scales slightly when hovered


craig-murray:
@thomasthesecond  How/where can I UX review this work and progress?

PR #153, opened 2017-03-03
Begin moving settings.json into styles directory

thomasthesecond:
As we continue to define a set of styles for LP apps, we’ll move away from the settings.json file and break styles into smaller modules.

These styles have been defined in the front-end framework.


jcreamer898:
Very cool.

PR #154, opened 2017-03-06
Create Toast notification component

### LINKED ISSUE #150: Toast
#### User Story
As a user I want a toast notification so I can notify the user of events

#### Acceptance Criteria
* It should fade in at the bottom of the page

thomasthesecond:
Closes #150

### DIFF
--- a/src/components/toast/index.jsx
+++ b/src/components/toast/index.jsx
@@ -1,0 +1,84 @@
+import React, { PropTypes } from "react";
+import radium from "radium";
+import colors from "../../styles/colors";
+import timing from "../../styles/timing";
+import { fontWeightMedium, fontSizeUppercase } from "../../styles/typography";
+import Icon from "../icon";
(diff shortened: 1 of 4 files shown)


src/components/toast/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+    className="Toast"
+    style={[
+      styles.container.default,
+      visible && styles.container.visible,
+      (!visible && direction === "bottom") && styles.container.invisibleBottom,
+      (!visible && direction === "top") && styles.container.invisibleTop,
+      style,
+    ]}
+  >
+    <Checkmark style={styles.icon} />
```
nickfranciosi: @thomasthesecond should this be passed in by a prop maybe? I assume that the Toast should also be able to handle an error state. So we might want to swap icons and colors based on error or success

thomasthesecond: Good idea thanks

PR #155, opened 2017-03-06
Create AvatarUpload component

thomasthesecond:


src/components/avatarUpload/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+    };
+
+    this.onChange = this.onChange.bind(this);
+  }
+
+  onChange(event) {
+    event.preventDefault();
+
+    if (this.input.files && this.input.files[0]) {
+      const reader = new FileReader();
```
nickfranciosi: according to MDN I don't think this will work in safari. https://developer.mozilla.org/en-US/docs/Web/API/FileReader. 

Might need a polyfill or fallback

thomasthesecond: I was  going off of caniuse. I'll look for a polyfill

thomasthesecond: Safari has basic support since 6.0. Tested and works. IE looks to have basic support since 10, I'll test.

thomasthesecond:
Safari and IE 10 are 👍

PR #172, opened 2017-03-21
Update profile header

thomasthesecond:
- Rename from "UserProfileHeader"
- Update props to reflect design; remove `subtitle`, add `intro`
- Add `interests` prop
- Add `alignment` prop for desktop, mobile layouts

Depends on #171


thomasthesecond:
image

image

jcreamer898:
So, I realize we don't yet have a profile app to play in, but should this kind of a component live in backpack? Just curious about your thoughts.

thomasthesecond:
I'm starting to think that yes, these kind of components should exist in Backpack. I think it would help us if we started using the [atomic design](http://bradfrost.com/blog/post/atomic-web-design/) mentality when building and describing our components, which describes five kinds of components:

- Atoms
- Molecules
- Organisms
- Templates
- Pages

I think Backpack should host atoms, molecules and organisms. Our components will be more maintainable, we'll always know where to look to find them, and they'll be available to all apps (and partners) at any time. It will also be nice to see the entire scope of our components in one place.

jcreamer898:
@thomasthesecond ❤️

PR #182, opened 2017-03-28
Refactor NewsArticleAuthor to use flexbox for more responsive layout

thomasthesecond:


thomasthesecond:
https://github.com/lonelyplanet/wp-travelnews-frontend/issues/81

src/components/newsArticleAuthor/index.jsx
```diff
@@ -1,22 +1,46 @@
...
+  },
+
   author: {
-    display: "block",
+    alignItems: "baseline",
+    display: "flex",
+    flexDirection: "row",
+    flexWrap: "wrap",
+    justifyContent: "flex-start",
+    marginTop: "17px",
```
nickfranciosi: my only comment would be that UX/Design seems to be wanting vertical distances that are multiples of 8 so `marginTop: 18px`. Surely there will be instances where this won't be possible or will change but I thought I would bring it up.

src/components/newsArticleAuthor/index.jsx
```diff
@@ -1,22 +1,46 @@
...
+  },
+
+  authorName: {
+    marginRight: "8px",
+    marginTop: "10px",
```
nickfranciosi: also here, but not necessarily needed to change

thomasthesecond: I use 10px here to maintain the spacing between flex elements. Otherwise, there's a noticeable shift when the component breaks down.

nickfranciosi: 👍

PR #192, opened 2017-04-03
add optional target='_blank' switch

laserchicken:
related p/r: https://github.com/lonelyplanet/copilot/pull/275

### DIFF
--- a/src/components/moreLink/index.jsx
+++ b/src/components/moreLink/index.jsx
@@ -65,6 +65,7 @@ function MoreLink({
   isNested,
   style,
   arrowDirection,
+  openNewTab,
 }) {
   let Element = "";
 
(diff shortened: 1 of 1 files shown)


src/components/moreLink/index.jsx
```diff
@@ -99,6 +100,7 @@ function MoreLink({
         style,
       ]}
       href={href}
+      {...(openNewTab && { target: "_blank" })}
```
thomasthesecond: A couple of things:

1. I'm of the opinion that we should use standard HTML props when possible. So I would recommend making the prop `target` and then allow `oneOf(["_blank", "_parent", "_self", "_top"])`
2. This should only apply when there is an `href`. If you notice, if an `onClick` is passed in, the element becomes a `button` and `target` does not apply. So you would have `(target && href) && target`. I don't think there's any reason to use the spread `...` operator here.

PR #213, opened 2017-04-10
add profile settings

nickfranciosi:
reference https://github.com/lonelyplanet/dotcom-profile/issues/24

- Adds UI for user profile settings blocks
- Separates existing checkbox ui from state that controls active logic
- Creates helper heightExpander component to make textArea height as tall as its content
- Creates ToggleController to wrap any component that needs simple boolean state management
- Fixes some issues radium has with `padding` shorthand


src/components/settingBlockAccordion/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+    height: 0,
+    maxHeight: 0,
+    overflow: "hidden",
+    transition: `max-height ${timing.slow} ease-in-out`,
+  },
+  accordionOpen: {
+    paddingBottom: "16px",
+    height: "auto",
+    maxHeight: "500vh",
+    overflow: "auto",
```
thomasthesecond: I'm seeing a flash of the scrollbar when the accordion opens. You don't need this here.

I think what you should have is this:

```
accordion: {
  overflow: "hidden",
  maxHeight: 0,
  transition: `max-height ${timing.slow} ease-in-out`,
},

accordionOpen: {
  paddingBottom: "16px",
  maxHeight: "500vh",
},
```

```
<div
  className="AccordionContent"
  style={[styles.accordion, expanded && styles.accordionOpen]}
>
  {children}
</div>
```

src/components/form/styles.js
```diff
@@ -93,11 +108,11 @@ const styles = {
...
-      borderBottom: `1px solid rgba(${rgb(color.lightBlue)}, 0.30)`,
-      ":focus": outline(),
-      error: {
-        borderBottom: `1px solid ${color.red}`,
-      },
+      borderBottom: 0,
+      // ":focus": outline(),
+      // error: {
+      //   borderBottom: `1px solid ${color.red}`,
+      // },
```
nickfranciosi: @thomasthesecond I forgot to my push my last update up. this is resolved

src/components/form/textarea.jsx
```diff
@@ -14,6 +15,8 @@ function TextArea({
   size,
   theme,
   onChange,
+  customStyles,
```
nickfranciosi: @thomasthesecond this is the already existing TextArea component. I was trying to keep it backwards compatible with anywhere that might be using it. Should I still update this?

src/components/settingBlock/settingBlockWrapper.jsx
```diff
@@ -0,0 +1,43 @@
...
+    ]}
+  >
+    {children}
+  </div>
+);
+
+SettingBlockWrapper.propTypes = {
+  children: PropTypes.element.isRequired,
+  error: PropTypes.bool,
+  hasAction: PropTypes.bool,
```
nickfranciosi: yes

nickfranciosi:
@thomasthesecond as for the components:

settingBlockAccordion
settingBlockCheckbox
settingBlockInput
settingBlockTextarea

I was trying to follow the same paradigm you used for cards. Do you not think this is a similar scenario. settingBlock is the base and is built from several smaller components. Then you have several different types of settingBlocks that are using the same smaller base component. just like there is a card component but also a cardVideo and cardBase.  Let me know you think

src/components/settingBlockAccordion/index.jsx
nickfranciosi: @thomasthesecond I had to change the transition from `max-height` to `all` to get this to work in both directions. Not sure why but if I just animated the `max-height` it was super janky when the accordion was closing. Any thoughts?

thomasthesecond:
Ah ok, I see. That makes sense, regarding the settingBlock components.

thomasthesecond:
In regards to the accordion, did the accordion we have not work in this instance? Seems like we should have one accordion component. But the accordion component we already have probably needs some work in order to be more flexible. I think AccordionItem should probably be split into two components, AccordionHeader and AccordionContent.

http://lonelyplanet.github.io/backpack-ui/?knob-Heading=High%20Sierra%20routes%20with%20Ken%20Walker%20Smith&knob-Bullets%5B0%5D=On%20The%20Road&knob-Bullets%5B1%5D=E.01&knob-Video%20runtime=129365&knob-Image%20source=https%3A%2F%2Flonelyplanetwp.imgix.net%2F2016%2F09%2FLPT0414_063-2225e4dcf106.jpg%3Ffit%3Dmin%26q%3D40%26sharp%3D10%26vib%3D20%26w%3D1470&knob-URL=%23&selectedKind=Accordion&selectedStory=Default&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel

src/components/form/styles.js
thomasthesecond: 👍

src/components/form/textarea.jsx
thomasthesecond: Ah, ok. I think we can leave it for now.

src/components/settingBlock/settingBlockWrapper.jsx
thomasthesecond: 👍

src/components/settingBlockAccordion/index.jsx
thomasthesecond: Good catch, I'm seeing it animates open but not closed. I think it has to do with the value of maxHeight.

If maxHeight is really tall like it is, and the actual content height is a fraction of the maxHeight, then the animation isn't going to look right. I think with 500vh and the speed of 200ms, it has to go a long way in a short time, so it's going to appear to collapse or expand without animating. It is, but it's too fast. I tried maxHeight 100px and I can see it animate open and closed. So it seems the best way is to calculate the actual height of the content for the animation to run smoothly. However, not sure the best way to do this without JS.

I also think the padding on the bottom causes a jump. The `AccordionContent` div shouldn't have the padding, but its content inside should, which in this case is `TagList`, but I would just wrap the content in a div with the padding.

```
<div className="AccordionContent" style={{...}}>
  <div style={{ paddingBottom: "16px" }}>
    {children}
  </div>
</div>
```

nickfranciosi: I added transition for just `max-height` and `padding-bottom` and it seems to work now actually -> `transition: max-height ${timing.slow} ease-in-out, padding-bottom ${timing.slow} ease-in-out`,

nickfranciosi:
I agree that it would be ideal to use one accordion everywhere, I just am not sure that the content/look of the two we have here would be easy to match up. I think we should definitely keep them purely presentational and all logic used to expand/close can be reused between accordions. I think that we  probably will use the accordion you built for any content actually on a page, while the one I just built seems to be more apt for forms, though it isn't every day you see an accordion in a form 😉 .

nickfranciosi:
@thomasthesecond I pushed up the changes we talked about. I got rid of the potential of nesting buttons in the accordion. Let me know if you see anything else

PR #234, opened 2017-04-19
Update react-slick to 0.14.10

thomasthesecond:
Attempt to fix failing production build due to recent changes to
dependencies in the react-slick package

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "react-recaptcha": "^2.2.5",
     "react-router": "^2.8.1",
     "react-scroll": "^1.4.4",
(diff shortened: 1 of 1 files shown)


thomasthesecond:
Guess that didn't fix it

thomasthesecond:
Failing build seems to be caused by enquire.js 2.1.6

thomasthesecond:
The latest version of react-slick 0.14.10 includes enquire 2.1.6 which
adds `window` as a global var and breaks the production build. 0.14.4
is the last known version that works, so this prevents the patch version
from updating every time a new build is started.

PR #245, opened 2017-04-21
feat: add taglist limiter

nickfranciosi:
adds wrapper around tagList to limit number of visible tags

### DIFF
--- a/src/components/tagList/index.jsx
+++ b/src/components/tagList/index.jsx
@@ -15,23 +15,56 @@ const styles = {
   },
 };
 
-const TagList = ({ children, rows, style }) => (
-  <div
-    className="TagList"
-    style={[
-      styles.container,
-      { maxHeight: `${rowHeight * rows}px` },
-      style,
-    ]}
-  >
-    {React.Children.map(children, (child, index) => (
-      React.cloneElement(child, {
-        key: index,
-        style: styles.tag,
-      })
-    ))}
-  </div>
-);
+class TagList extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      expanded: false,
+    };
+
+    this.expand = this.expand.bind(this);
+    this.shouldShowExpand = this.shouldShowExpand.bind(this);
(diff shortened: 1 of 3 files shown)


src/components/tagList/index.jsx
```diff
@@ -24,12 +24,13 @@ const TagList = ({ children, rows, style }) => (
...
-        style: styles.tag,
-      })
-    ))}
+    <Style
+      scopeSelector=".TagList"
+      rules={{
+        ".Tag": styles.tag,
+      }}
+    />
+    {children}
```
thomasthesecond: What is the reason for this change?

stories/index.jsx
```diff
@@ -2140,8 +2141,24 @@ storiesOf("Tag list", module)
...
+        <Tag href="#">World</Tag>,
+        <Tag href="#">Asia & the Pacific</Tag>,
+        <Tag href="#">Europe</Tag>,
+        <Tag href="#">Middle East & Africa</Tag>,
+        <Tag href="#">Editor’s pick</Tag>,
+        <Tag href="#">Brains</Tag>,
+        <Tag href="#">Mustard</Tag>,
+        <Tag href="#">Noodles</Tag>,
+        <Tag href="#">Ed’s pick</Tag>,
+      ]}
```
thomasthesecond: Can these `Tag`s be passed as children instead?

src/components/tagListLimiter/index.jsx
```diff
@@ -0,0 +1,53 @@
+import React, { PropTypes } from "react";
+import TagList from "../tagList";
+import Tag from "../tag";
+
+class TagListLimiter extends React.Component {
```
thomasthesecond: What do you think about extending `TagList` to accommodate this behavior? What if we just passed in the limit prop to `TagList` and added state? And if you don't pass a limit it just shows them all.

nickfranciosi:
@thomasthesecond I moved this logic into the TagList. Thanks!

PR #246, opened 2017-04-24
Updated Segment component style logic so custom styling can always be applied

harterc1:


harterc1:
These changes are no longer needed.  The style attributes i was trying to override negated the purpose of the Segment component in the first place -- so i just stopped using the Segment component.

PR #266, opened 2017-05-01
Create AvatarMarker component

thomasthesecond:
image


thomasthesecond:
Should help with lonelyplanet/dotcom-profile#2

thomasthesecond:
It doesn't, however it could with https://github.com/tuchk4/storybook-readme

There's also this https://github.com/storybooks/storybook/tree/master/packages/addon-notes but I don't like how you write the notes inline.

I've seen READMEs added to other component libraries and thought I'd give it a try. We don't have to make it a thing if we don't want. However, I think sometimes when you have a component name like `AvatarMarker`, its purpose isn't entirely clear from the name. So I added a README to help with that in this case. Naming is hard.

nickfranciosi:
For sure. I love the concept. Im down

PR #296, opened 2017-05-12
update: Update typo in newsletter component

katelynp43:
Replaced hardcoded text with default props. Updated the default text and fixed typo.


thomasthesecond:
One thing I've realized is that standard-version only accepts the following tags:

- `fix`
- `feat`
- `BREAKING CHANGE`

It basically ignores anything else when updating the changelog.

so this would be a `fix: `. You can easily change it before you merge, I can show you.

PR #309, opened 2017-05-25
Create Multi-step component

### LINKED ISSUE #300: Create multi-step modal
#### Acceptance criteria

* Accepts multiple steps as children that are other React components
* Handles state to track current step
* Contains "previous" and "next" methods that navigate between steps
* Contains logic to display previous and next steps
* Nice to have: animate between steps

thomasthesecond:
Closes #300

### DIFF
--- a/src/components/multiStep/multiStep.jsx
+++ b/src/components/multiStep/multiStep.jsx
@@ -10,13 +10,7 @@ const MultiStep = ({ children, currentStep, className, id, style }) => (
     style={style}
   >
     {Children.map(children, (child, index) => (index + 1) === currentStep && (
-      <div
-        className={className ? `${className}-step-${index}` : `MultiStep-step-${index}`}
-        id={id ? `${id}-step-${index}` : null}
-        key={`step-${index}`}
-      >
-        {child}
-      </div>
+      child
     ))}
   </div>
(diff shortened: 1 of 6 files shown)


src/components/multiStep/multiStep.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const MultiStep = ({ children, currentStep, className, id, style }) => (
+  <div
+    className={cn("MultiStep", className)}
+    id={id}
+    style={style}
+  >
+    {Children.map(children, (child, index) => (index + 1) === currentStep && (
+      <div
+        className={className ? `${className}-step-${index}` : `MultiStep-step-${index}`}
```
egdelwonk: Could you default the `className` prop to `'MultiStep'`, and then always use `className` here?

thomasthesecond:
Good call. I can pass that method down so it's available.

src/components/multiStep/multiStep.jsx
thomasthesecond: I always want the `MultiStep` classname to be available and I intended for the `className` prop to add a modifier classname. Maybe the children don't need classnames or IDs so I could simplify the code some.

PR #311, opened 2017-05-25
feat: pass through social actions to loginModal

nickfranciosi:
BREAKING CHANGE: this addition breaks the older implemention of having the links live inside the component

### DIFF
--- a/src/components/modalContent/modalContentSocialAuth.jsx
+++ b/src/components/modalContent/modalContentSocialAuth.jsx
@@ -45,8 +45,8 @@ const styles = {
   },
 };
 
-const disclaimer = `If you sign up with Twitter, Facebook or Google, weâ€™ll automatically import your
-  profile information. Weâ€™ll never post without your permission. Alternatively sign in with a
+const disclaimer = `If you sign up with Twitter, Facebook or Google, we’ll automatically import your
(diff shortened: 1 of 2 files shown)


thomasthesecond:
Are you thinking of bumping to version 5.0? I had a list of breaking changes I wanted to merge all at once here, but not sure how reasonable it would be to get them all complete in a timely manor.

https://github.com/lonelyplanet/backpack-ui/milestone/2

nickfranciosi:
@thomasthesecond technically we would for this, but I don't think anything is consuming this component, but if you have other things to add. I am not in a hurry to have it live, I can work with it locally

thomasthesecond:
Ok cool. Most of those changes I want to make are pretty small. The button refactor will have the largest effort, but I'm actually almost done with that. So maybe we can make a breaking change release today or tomorrow?

thomasthesecond:
Also I think for standard-version to pick this up as a breaking change, you would replace `feat:` with `BREAKING CHANGE:` in the subject line.

`BREAKING CHANGE: pass through social actions to loginModal`

thomasthesecond:
We might not need this update any more. The components in https://github.com/lonelyplanet/backpack-ui/pull/322 will eventually deprecate the `ModalContentSocialAuth` component.

PR #324, opened 2017-05-30
Create color utility function to create overlay colors

### LINKED ISSUE #206: Make bgOverlay color 30% black
`bgOverlay` is currently solid black. It should be 30% black.

This change will break the current implementation of `bgColor` because its value is changing from hex to rgba and `bgColor` is currently being used with the `rgba` function to adjust opacity.

thomasthesecond:
Closes #206


thomasthesecond:
Don't merge yet.

PR #325, opened 2017-05-31
fix: update avatar src with prop change

nickfranciosi:


src/components/checkbox/checkboxComponent.jsx
```diff
@@ -65,102 +65,96 @@ const styles = {
...
+  onClick,
+  id,
+  value,
+  size,
+  name,
+  label,
+  rounded,
+  style,
+}) => {
+  const textPadding = {
```
ruebel: Shouldn't this be declared outside of the render function so it doesn't have to be recreated every time you render?

PR #356, opened 2017-06-28
fix: change navigation zindex to 99 to stay behind public profile dropdowns

shawncothran:
Closes lonelyplanet/dotcom-profile#104...

### DIFF
--- a/src/components/navigation/navigation.jsx
+++ b/src/components/navigation/navigation.jsx
@@ -11,7 +11,7 @@ const styles = {
     position: "relative",
     textAlign: "center",
     transform: "translateZ(0)",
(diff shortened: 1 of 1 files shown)


src/components/navigation/navigation.jsx
```diff
@@ -11,7 +11,7 @@ const styles = {
     position: "relative",
     textAlign: "center",
     transform: "translateZ(0)",
-    zIndex: zIndex.globalHeader,
+    zIndex: zIndex.globalHeader -1,
```
thomasthesecond: This causes a linting error. You'll need to put a space before `1`

`zIndex.globalHeader - 1`

PR #359, opened 2017-06-30
update Typeahead component

laserchicken:


src/components/typeahead/index.jsx
```diff
@@ -43,8 +37,6 @@ const Typeahead = ({
           marginTop: 0,
           overflow: "hidden",
           paddingLeft: 0,
-          position: "absolute",
-          top: "auto",
```
thomasthesecond: What was the reason for removing these styles? Was it causing layout issues?

laserchicken: yes this is how it looked:
screen shot 2017-07-05 at 20 27 15
without `position: "absolute"` we will have normal flow of elements:
screen shot 2017-07-05 at 20 35 05

thomasthesecond: 👍 cool, thanks

PR #361, opened 2017-07-03
fix: add max-width to disclaimer text for wrapping in IE

shawncothran:
also auto hide scrollbars on container


nickfranciosi:
@shawncothran usually wait until you get a 👍 before merging 😄

PR #367, opened 2017-07-07
Delete deploy scripts from package.json

thomasthesecond:
Because master is now a protected branch, there is no use for the
deploy scripts anymore. These scripts combined the versioning, pushing
tags and publishing, however we are no longer able to perform these
actions all at once.


thomasthesecond:
I think I'll leave these scripts in for admins.

PR #373, opened 2017-07-13
Add HOC to detect outside clicks

thomasthesecond:
Enhances a React component with a Higher Order Component that provides click outside detection


thomasthesecond:
Not sure if it's the "right" way, but it gets around using `ReactDOM.findDOMNode`.

PR #374, opened 2017-07-14
Allow HTML content in TextBodySmall

thomasthesecond:
https://github.com/lonelyplanet/dotcom-profile/issues/129


thomasthesecond:
Yeah good points. I also just realized that it's markdown content in the issue above, not HTML, so this wouldn't work for legacy users I don't think.

thomasthesecond:
I'll hold off on this for now and see how @bpotterbaum wants to proceed.

PR #380, opened 2017-07-27
Update bookmark icons

shawncothran:


src/components/icon/index.jsx
```diff
@@ -331,6 +331,20 @@ exports.DiamondLogo = props => (
...
+  <Icon {...props}>
+    <path d="M5.333 0.667h21.333v30.667l-10.524-5.333-10.809 5.333v-30.667zM24 3.333h-16v23.711l8.158-4.025 7.842 3.974v-23.659zM24 4.667v14.667h2.667v-14.667h-2.667z" />
+    <path d="M26.667 10.667v-3.333h-2.667v3.333h-3.333v2.667h3.333v3.333h2.667v-3.333h3.333v-2.667h-3.333z" />
+  </Icon>
+);
+
+exports.ListBookmarked = props => (
+  <Icon {...props}>
+    <path d="M5.333 0.667h21.333v30.667l-10.524-5.333-10.809 5.333v-30.667zM24 3.333h-16v23.711l8.158-4.025 7.842 3.974v-23.659zM24 4.667v14.667h2.667v-14.667h-2.667z" />
+    <path fill="#16c98d" d="M20.943 9.724l-1.885 1.885 4.943 4.943 7.609-7.609-1.885-1.885-5.724 5.724z" />
```
thomasthesecond: Since we're importing `colors` here, can we we `colors.accentGreen` here?

These icons should replace `Bookmark` and `BookmarkOutline`; we don't want multiple versions of these icons.
* Delete `BookmarkOutline`
* `ListBookmark` becomes `Bookmark`
* `ListBookmarked` becomes `BookmarkActive`

PR #387, opened 2017-07-27
Create UserListAuthor component

### LINKED ISSUE #379: Create author component for user lists
It should use the Avatar and AuthorName components.

Avatar with is 22 pixels but this does not fall on the 8pt grid; this will need to be addressed. 24 pixels is an accepted size for the Avatar component. See https://github.com/lonelyplanet/backpack-ui/blob/master/src/components/avatar/index.jsx#L76

image

Sketch file: https://www.dropbox.com/work/Digital%20Product/user%20lists/bp?preview=lp_lists_aligned_bp.sketch

_image not to scale_

thomasthesecond:
Closes #379

### DIFF
--- a/src/components/userListAuthor/index.jsx
+++ b/src/components/userListAuthor/index.jsx
@@ -1,0 +1,51 @@
+import React from "react";
+import PropTypes from "prop-types";
+import radium from "radium";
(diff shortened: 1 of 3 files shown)


thomasthesecond:
Comment

Issue #383, opened 2017-07-27
Create heading for user list


thomasthesecond:
![image](https://user-images.githubusercontent.com/1479563/28678230-70bfd854-72b5-11e7-8545-7f852b1c2499.png)

Sketch file: https://www.dropbox.com/work/Digital%20Product/user%20lists/bp?preview=lp_lists_aligned_bp.sketch

_image not to scale_

shawncothran:
<img width="520" alt="screen shot 2017-08-01 at 4 32 37 pm" src="https://user-images.githubusercontent.com/14968386/28848071-33931e2a-76d7-11e7-9cb7-1ef5bcad7544.png">
Desktop Version

PR #396, opened 2017-08-07
feat: Create ListHeader component (#383)

shawncothran:


src/components/listHeader/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+    [`@media (min-width: ${mq.min["768"]})`]: {
+      fontSize: `${fontSizeHeading6}px`,
+      letterSpacing: 0,
+    },
+  },
+};
+
+const ListHeader = ({
+  href,
```
nickfranciosi: @shawncothran maybe name `href` and `src` to more clearly show that they are the users image src and the link to the users profile. Otherwise looks great 👍

egdelwonk: ☝️ I agree.

Issue #377, opened 2017-07-27
Create thumbnail component for user list item


thomasthesecond:
Default (no image)*:
![image](https://user-images.githubusercontent.com/1479563/28676928-a6f3574c-72b1-11e7-8839-b58ed0c230b7.png)

Add*:
![image](https://user-images.githubusercontent.com/1479563/29040763-df30296a-7b75-11e7-954d-0ab6bd36ef97.png)

Image*:
![image](https://user-images.githubusercontent.com/1479563/28677042-f848ab1a-72b1-11e7-8aaf-729d7c166f8c.png)

Note the masking on the bars on the right side of the image.

Sketch file: https://www.dropbox.com/work/Digital%20Product/user%20lists/bp?preview=lp_lists_aligned_bp.sketch

_* not to scale_

PR #397, opened 2017-08-07
Create thumbnail image for user list

thomasthesecond:
#377


thomasthesecond:
Updated to include icons. Defaults to the list icon if no `src` is passed.

PR #407, opened 2017-08-08
feat: Add ListEntry component, header border and priceRange size (#348)

shawncothran:


src/components/listEntry/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+import { rgba } from "../../utils/color";
+
+const styles = {
+  borderBottom: `1px solid ${colors.borderPrimary}`,
+  padding: "16px",
+
+  [`@media (min-width: ${mq.min["768"]})`]: {
+    padding: "24px 0",
+  },
+
```
thomasthesecond: Can we wrap these styles in a `container` key? So instead of using `styles` and getting them all, we can just get the ones we need with `styles.container`. This is generally the approach I follow; grouping styles by element.

src/components/listEntry/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+  name,
+  category,
+  city,
+  note,
+  value,
+  style,
+}) => (
+  <article
+    className="ListEntry"
+    style={[styles, style]}
```
thomasthesecond: Then here you would have `styles.container`

src/components/listEntry/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+  >
+    <PriceRangeLabel
+      value={value}
+      style={styles.value}
+    />
+
+    <h2
```
thomasthesecond: Can we use the Heading component here?

`import { Heading } from "../../components/text";`

src/components/listEntry/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+      className="ListEntry"
+      style={[styles.name, style]}
+    >
+      {name}
+    </h2>
+
+    <h3 style={styles.category}>
+      {category} in {city}
+    </h3>
```
thomasthesecond: Semantically, I think this can be a `div` or a `p`

thomasthesecond: On second thought, I think we can use the CategoryLabel component here.

src/components/listEntry/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+  </article>
+);
+
+ListEntry.propTypes = {
+  name: PropTypes.string.isRequired,
+  category: PropTypes.objectOf(PropTypes.array).isRequired,
+  city: PropTypes.objectOf(PropTypes.array).isRequired,
+  note: PropTypes.bool.isRequired,
+  value: PropTypes.oneOf(["$", "$$", "$$$"]).isRequired,
+  style: PropTypes.objectOf(PropTypes.object),
```
thomasthesecond: We have a custom prop type we use for `style`

`import propTypes from "../../utils/propTypes";`

and then `style: propTypes.style,`

src/components/listEntry/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+      {note}
+    </p>
+  </article>
+);
+
+ListEntry.propTypes = {
+  name: PropTypes.string.isRequired,
+  category: PropTypes.objectOf(PropTypes.array).isRequired,
+  city: PropTypes.objectOf(PropTypes.array).isRequired,
+  note: PropTypes.bool.isRequired,
```
thomasthesecond: Based on how `note` is used above, I think this type should be a `string`.

src/components/listEntry/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+      {name}
+    </h2>
+
+    <h3 style={styles.category}>
+      {category} in {city}
+    </h3>
+
+    <p style={styles.note}>
+      {note}
+    </p>
```
thomasthesecond: I think we can use the [TextAccent](https://github.com/lonelyplanet/backpack-ui/blob/master/src/components/text/textAccent.jsx) component here

import { TextAccent } from "../../components/text";

It applies this mixin: https://github.com/lonelyplanet/backpack-ui/blob/master/src/utils/typography.js#L148-L155

src/components/listEntry/index.jsx
thomasthesecond: Also, not sure if this should be required. Probably just name, category and city.

shawncothran: total oversight! You're right.

PR #433, opened 2017-08-16
feat: add desktop styles to bookmark button

shawncothran:


src/components/bookmarkButton/index.jsx
```diff
@@ -1,10 +1,25 @@
...
 const BookmarkButton = ({ onClick, marked, iconType, style }) => {
+  const styles = {
+    [`@media (min-width: ${mq.max["768"]})`]: {
+      boxShadow: "none",
+      display: "flex",
+      fontSize: fontSizeBodyArticle,
+      marginBottom: "40px",
+      marginLeft: "auto",
+      marginRight: "auto",
+      marginTop: "-21px",
```
thomasthesecond: I think margins should be controlled by the app that imports the component. These values are specific only to the POIs implementation. The component should be able to be viewed on its own without layout or position implementation.

PR #438, opened 2017-08-17
fix: remove positioning related styles from bookmark button

shawncothran:


thomasthesecond:
Merged to get this fix into 4.32.0

PR #443, opened 2017-08-18
fix: remove exif data from avatar uploader

nickfranciosi:
photos uploaded with exif data currently are retaining gps data which could be  huge privacy concern. 

this removes the exif data and converts png's to jpgs by writing the image data to a canvas and then reading it back out as a dataUri to pass to the uploader


nickfranciosi:
@thomasthesecond - yeah I was trying to abstract this logic out but so much is dependent on the `onload` function firing. I think there is definitely a way to pull it out though, but thought it might complicate it more for right now

PR #451, opened 2017-08-24
Add topChoice prop, make city prop not required in bookmark list entry

thomasthesecond:


thomasthesecond:
I could just add a conditional around `in {city}` so if there's no place it won't look weird. Might be good to do that any way.

thomasthesecond:
Updated

PR #503, opened 2017-09-26
fix: show modal title via bool prop

shawncothran:
Allow modal title to remain visible if showTitle prop is passed, otherwise "display: none" in desktop media query

### DIFF
--- a/src/components/modal/index.jsx
+++ b/src/components/modal/index.jsx
@@ -63,10 +63,6 @@ const styles = {
 
   title: Object.assign({}, {
     display: "inline-block",
-
-    [`@media ${largeMQ}`]: {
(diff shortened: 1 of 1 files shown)


src/components/modal/index.jsx
```diff
@@ -188,7 +185,16 @@ function ModalComponent({
         }
 
         {title &&
-          <span style={styles.title}>
+          <span
+            style={[
+              styles.title,
+              !showTitle && {
+                [`@media (min-width: ${mq.min["768"]})`]: {
```
thomasthesecond: We can use ```[`@media ${largeMQ}`]``` here from line 67 above

PR #507, opened 2017-09-27
fix: disable scroll lock off after navigate

nickfranciosi:
currently if a user navigates away from a page while a modal is open, the scroll lock on the html stays in place so you can't scroll

this fixes the issue by listening to navigation and unMount events to remove the fixed window styles

### DIFF
--- a/src/components/modal/wrapper.jsx
+++ b/src/components/modal/wrapper.jsx
@@ -11,6 +11,15 @@ class ModalWrapper extends React.Component {
     };
 
     this.toggleOpen = this.toggleOpen.bind(this);
+    this.handleNavigateAway = this.handleNavigateAway.bind(this);
+  }
+
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/components/modal/wrapper.jsx
```diff
@@ -25,6 +34,11 @@ class ModalWrapper extends React.Component {
     });
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  handleNavigateAway() {
```
thomasthesecond: Could this be made a static method? Then you wouldn't have to disable the lint rule.

nickfranciosi: if I make it a static method then I have to call it in `componentDidMount` then the `this` keyword is not being used in that function and I get the same eslint error

thomasthesecond: Ah ok

PR #524, opened 2017-10-13
fix: make textarea controlled input

laserchicken:


src/components/textarea/index.jsx
```diff
@@ -76,6 +78,13 @@ class Textarea extends React.Component {
     }
   }
 
+  handleChange(e) {
+    if (this.props.onChange) {
```
thomasthesecond: Can we also check `typeof this.props.onChange === "function"`?

laserchicken:
sure, updated ☝️

thomasthesecond:
Thanks. I'll make a new release.

PR #525, opened 2017-10-13
feat: add reveal with flyout to bookmarkListMenu

shawncothran:
For: https://github.com/lonelyplanet/dotcom-profile/issues/312

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,11 +76,11 @@
     "react": "^15.0.0 || ^16.0.0"
   },
   "devDependencies": {
+    "@types/react": "^16.0.5",
     "@storybook/addon-actions": "^3.1.9",
     "@storybook/addon-knobs": "^3.0.0",
     "@storybook/addon-links": "^3.0.0",
     "@storybook/react": "^3.0.1",
-    "@types/react": "^16.0.5",
     "babel-cli": "^6.14.0",
     "babel-eslint": "^6.1.2",
     "babel-plugin-transform-decorators-legacy": "^1.3.4",
--- a/src/components/bookmarkListMenu/index.jsx
+++ b/src/components/bookmarkListMenu/index.jsx
@@ -4,8 +4,10 @@ import radium from "radium";
 import cn from "classnames";
 import Flyout from "../flyout";
 import IconButton from "../iconButton";
+import IconRevealButton from "../iconRevealButton";
 import colors from "../../styles/colors";
 import timing from "../../styles/timing";
+import iconFromString from "../../utils/icon";
 import propTypes from "../../utils/propTypes";
 import clickOutside from "../../hoc/clickOutside";
 
@@ -92,6 +94,7 @@ class BookmarkListMenu extends React.Component {
       iconLabel,
       id,
       className,
+      reveal,
       style,
     } = this.props;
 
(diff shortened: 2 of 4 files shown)


src/components/bookmarkListMenu/index.jsx
```diff
@@ -144,13 +159,11 @@ class BookmarkListMenu extends React.Component {
 BookmarkListMenu.propTypes = {
   innerRef: PropTypes.func.isRequired,
   children: PropTypes.element.isRequired,
-  iconName: PropTypes.oneOf([
-    "Ellipsis",
-    "Share",
-  ]).isRequired,
+  iconName: PropTypes.string.isRequired,
```
thomasthesecond: Can we keep `oneOf` here? I don't think any of our icons should be used here.

package.json
```diff
@@ -76,10 +80,6 @@
     "react": "^15.0.0 || ^16.0.0"
   },
   "devDependencies": {
-    "@storybook/addon-actions": "^3.1.9",
-    "@storybook/addon-knobs": "^3.0.0",
-    "@storybook/addon-links": "^3.0.0",
-    "@storybook/react": "^3.0.1",
```
thomasthesecond: Storybook docs recommend these to be a devDependency

src/components/bookmarkListMenu/index.jsx
```diff
@@ -104,15 +110,24 @@ class BookmarkListMenu extends React.Component {
...
-          iconName={iconName}
-          label={iconLabel}
-          size={40}
-          owns={optionsId}
-          color={colors.textPrimary}
-          backgroundColor={colors.bgPrimary}
-        />
+        {reveal ?
+          <IconRevealButton
+            id={"poiId"}
```
thomasthesecond: Can this be generated dynamically based on the `id` prop?

thomasthesecond: ```id={`${id}-IconRevealButton`}```

src/components/bookmarkListMenu/index.jsx
```diff
@@ -144,13 +159,11 @@ class BookmarkListMenu extends React.Component {
...
   children: PropTypes.element.isRequired,
-  iconName: PropTypes.oneOf([
-    "Ellipsis",
-    "Share",
-  ]).isRequired,
+  iconName: PropTypes.string.isRequired,
   iconLabel: PropTypes.string.isRequired,
   id: PropTypes.string,
   className: PropTypes.string,
+  reveal: PropTypes.bool,
```
thomasthesecond: Can we add a default prop as false for `reveal`?

stories/index.jsx
```diff
@@ -484,8 +482,9 @@ storiesOf("Bookmark list menu", module)
   .add("Default", () => (
     <Center>
       <BookmarkListMenu
-        iconName={select("Icon name", ["Ellipsis", "Share"], "Ellipsis")}
+        iconName={text("Icon name", "Ellipsis")}
```
thomasthesecond: This would be reverted based on my comment above.

src/components/bookmarkListMenu/index.jsx
```diff
@@ -2,13 +2,18 @@ import React from "react";
...
 import Flyout from "../flyout";
 import IconButton from "../iconButton";
+import IconRevealButton from "../iconRevealButton";
 import colors from "../../styles/colors";
 import timing from "../../styles/timing";
+import iconFromString from "../../utils/icon";
 import propTypes from "../../utils/propTypes";
 import clickOutside from "../../hoc/clickOutside";
 
+const _ = { upperFirst };
```
thomasthesecond: May not need this if `iconName` prop type is reverted to `oneOf`

package.json
shawncothran: Oh ok, I changed them based on eslint

src/components/bookmarkListMenu/index.jsx
shawncothran: oversight

PR #530, opened 2017-10-18
Make sure profile header URL has `http://` in href

thomasthesecond:
We don't want to show `http://` in the UI, but the href needs
it. The bug was caused by the assumption that `http://` would
always be there. The fix is just to remove `http://` and
then add it back in the href so it will always be there.

### DIFF
--- a/src/components/profileHeader/index.jsx
+++ b/src/components/profileHeader/index.jsx
@@ -225,11 +225,15 @@ class ProfileHeader extends React.Component {
       }
     `;
 
-    const formattedWebsite = website ? website
-      .replace("https://", "")
-      .replace("http://", "")
-      .replace("www.", "")
-    : "";
+    const urlRegex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_+.~#?&//=]*)/; // eslint-disable-line max-len
+
+    const validateUrl = (url) => urlRegex.test(url);
+
+    const urlParser = (url) => {
+      const anchor = document.createElement("a");
+      anchor.href = url;
+      return anchor;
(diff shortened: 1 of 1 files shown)


src/components/profileHeader/index.jsx
```diff
@@ -274,12 +280,12 @@ class ProfileHeader extends React.Component {
...
                 ]}
               >
-                <a href={website} target="_blank" rel="noopener noreferrer">
-                  {website
-                    .replace("https://", "")
-                    .replace("http://", "")
-                    .replace("www.", "")
-                  }
+                <a
+                  href={`http://${formattedWebsite}`}
```
WunderBart: I think it would be safer to use build in parser like here: https://gist.github.com/jlong/2428561
Forcing `http` and removing `www` can cause troubles.

WunderBart: (if there's no protocol specified then we could add `http://`)

thomasthesecond: Good idea! I'll update this.

thomasthesecond: Running into an issue where if someone inputs "lonelyplanet.com", the parser interprets that as the `pathname`.

`http://url.of.the.profile.page/lonelyplanet.com`

I think we need better validation on the field itself on the profile edit page.

thomasthesecond: https://github.com/lonelyplanet/dotcom-profile/pull/402

thomasthesecond:
Added a parser and validator for the URL.

thomasthesecond:
Thought about this, but ... If someone adds a link to say, their personal website, and it's not https and they don't have SSL, not sure I should be forcing https.

jcreamer898:
Ahhh ok that's fine then.

PR #540, opened 2017-10-23
Updated cue point logic to respect all text tracks;  Video pixels will now be inserted into page when available

harterc1:


src/components/videoEmbed/index.jsx
```diff
@@ -192,59 +195,46 @@ class VideoEmbed extends Component {
...
-    while (index < tt.cues.length) {
-      const cue = tt.cues[index];
-      if (cue.text === "CODE") {
-        cues.push(cue);
-      }
-      index += 1;
-    }
-
-    return cues;
+  getActiveCues() {
```
jcreamer898: I think you can just use `map` here.

```js
getActiveCues() {
  const activeCues = this.player.textTracks().tracks_
    .map((tt) => tt.activeCues_)

  return activeCues;
}
```

PR #551, opened 2017-11-10
fix: cursor jumps out of inputs in modal

nickfranciosi:
fixes https://github.com/lonelyplanet/dotcom-profile/issues/407

### DIFF
--- a/src/components/modal/index.jsx
+++ b/src/components/modal/index.jsx
@@ -134,6 +134,7 @@ function ModalComponent({
       top: 0,
       WebkitOverflowScrolling: "touch",
       width: "100%",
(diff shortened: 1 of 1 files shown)


src/components/modal/index.jsx
```diff
@@ -143,7 +144,7 @@ function ModalComponent({
           top: "50%",
           width: desktopWidth,
           maxWidth: "1290px",
-          transform: "translateY(-50%)",
+          transform: "translate3d(0px, -50%, 0px) !important",
```
thomasthesecond: Can just use `0` here and above, don't need `px`

nickfranciosi:
@thomasthesecond thanks! should be resolved now

PR #560, opened 2017-11-20
Refactor SectionHeader component

thomasthesecond:
This PR makes a few changes to the SectionHeader component.

* Replace media queries with react-measure. This allows for the component’s size to dictate its responsive styles instead of the window’s size. In turn, the component has been converted from functional to stateful because the dimensions provided by react-measure need to be handled within the component’s state.
* Remove containing `header` element. There is no reason for an extra container around the Heading component, so it has been removed.
* Use new TextHeading component and typography styles
* Change divider from `hr` to a `span`. Ideally, the divider would be build purely in CSS using pseudo elements; this change mimics that idea. 1. A pseudo element lives inside of it’s parent element, so the divider markup has been moved inside of Heading; this also allows for the sizing to be relative to the Heading’s font size. 2. A pseudo element has no semantic value, neither does a `span` and since the divider is presentational, the `hr` provides semantic value we don’t want.
* Remove heading object prop; since the component itself is now a Heading, we don’t need a separate prop to pass in an object of settings for the Heading. Furthermore, the Heading now has it’s props predetermined, i.e., size, level, weight. Previously, the SectionHeading could include any number of heading props, but now I think it’s better to limit the component to be specifically configured how it was designed; an h2, using our fontSizeHeading2 (4 for mobile) and regular weight. This will allow the component to be rendered consistently.
* These updates should not break any current implementation, however, any heading settings passed in will not be applied.

### DIFF
--- a/src/components/sectionHeader/index.jsx
+++ b/src/components/sectionHeader/index.jsx
@@ -59,8 +59,6 @@ class SectionHeader extends React.Component {
     const { width } = this.state.dimensions;
     const { children, theme, style } = this.props;
 
-    console.log(width);
-
     return (
       <Measure
         bounds
--- a/src/components/sectionHeader/index.jsx
+++ b/src/components/sectionHeader/index.jsx
@@ -4,11 +4,7 @@ import radium from "radium";
 import Measure from "react-measure";
 import { Heading } from "../text";
 import colors from "../../styles/colors";
-import {
-  fontSizeHeading2,
-  fontSizeHeading4,
-  lineHeightHeading4,
-} from "../../styles/typography";
+import { fontSizeHeading2 } from "../../styles/typography";
 import propTypes from "../../utils/propTypes";
 
 const styles = {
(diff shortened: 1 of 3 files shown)


src/components/sectionHeader/index.jsx
```diff
@@ -1,94 +1,118 @@
...
   divider: {
-    width: "30px",
-    borderStyle: "solid",
-    borderWidth: "1px",
-    marginBottom: "32px",
+    display: "block",
+    height: "2px",
+    marginLeft: "auto",
+    marginRight: "auto",
+    marginTop: `${16 / fontSizeHeading2}em`,
```
nickfranciosi: @thomasthesecond to use `em` here do we also need to set the font-size of the parent container to be equal to the heading font-size? or are we not adjusting that margin based on the header font size?

would we use `rem` if want to just base it off the root size?

thomasthesecond: My reason for `em` here is to change the top margin and width of the divider to the same proportions whenever the heading font size is changed. Since on mobile, the font size gets smaller, so too will the margin and width of the divider.

The font size on the header is set with the `size={2}` prop, so that's why I didn't set it here.

nickfranciosi: sorry I think I missed this was nested in the `<Header />` element

thomasthesecond: No worries, thanks!

PR #564, opened 2017-11-24
fix: allow hiding modal header

laserchicken:
I need this for https://github.com/lonelyplanet/dotcom-profile/pull/449

### DIFF
--- a/src/components/modal/index.jsx
+++ b/src/components/modal/index.jsx
@@ -84,6 +84,7 @@ function ModalComponent({
   disableContentPadding,
   title,
   showTitle,
+  showHeader,
   className,
   children,
   style,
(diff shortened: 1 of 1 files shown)


thomasthesecond:
Though this has been merged, my feeling is this should be `hideHeader` instead of `showHeader`. I think we should always have boolean props default value be false, so using `hideHeader` would make this so.

Reason being, showing the header is the default behavior or the primary use case; generally we always want to show the header and there are only a few cases where we would want to hide it, so in those cases we can just pass `hideHeader`.

Another reason, the code is cleaner and easier to read. Compare:

`<Modal showHeader={false} />`

to

`<Modal hideHeader />`

The first version goes against how we've been handling boolean props in other components. In React, you don't need to specify `true` or `false` with boolean props. The prop is there (`true`) or it's not (`false`). Similar to HTML5 boolean attributes (`required`, `selected`, `hidden`, etc).

In the second example, we're explicitly stating what we want. I think it's easier to look at and comprehend what's going on.

These are things to keep in mind when creating boolean props.

PR #567, opened 2017-11-30
fix: use default hideHeader instead of showHeader

laserchicken:
request: https://github.com/lonelyplanet/backpack-ui/pull/564#issuecomment-347890858


laserchicken:
cc @thomasthesecond

PR #575, opened 2017-12-04
feat: add url parameter to toast action location

shawncothran:
reference: https://github.com/lonelyplanet/dotcom-profile/issues/445


src/components/toast/index.jsx
```diff
@@ -237,27 +244,29 @@ const Toast = ({
...
   direction: PropTypes.oneOf(["top", "bottom"]),
-  visible: PropTypes.bool,
-  affixed: PropTypes.bool,
   onClick: PropTypes.func,
   onClose: PropTypes.func,
-  title: PropTypes.string,
-  buttonLabel: PropTypes.string,
   style: propTypes.style,
+  title: PropTypes.string,
+  type: PropTypes.oneOf(["error", "info", "success", "warning"]).isRequired,
```
thomasthesecond: Can we keep required props at the top of the object?

src/components/toast/index.jsx
```diff
@@ -190,19 +191,19 @@ const Toast = ({
...
     <Style
-      scopeSelector=".Toast"
       rules={{
         mediaQueries: {
           "(prefers-reduced-motion)": {
             transform: "translateY(0) !important",
           },
         },
       }}
+      scopeSelector=".Toast"
```
thomasthesecond: Can we keep this above `rules`

src/components/toast/index.jsx
```diff
@@ -148,24 +148,25 @@ const messageTypes = {
...
   <div
-    className="Toast"
-    role="alertdialog"
+    aria-describedby="toastMessage"
     aria-hidden={!visible}
-    aria-live="assertive"
     aria-labelledby="toastTitle"
-    aria-describedby="toastMessage"
+    aria-live="assertive"
+    className="Toast"
```
thomasthesecond: `className` first please

src/components/toast/index.jsx
```diff
@@ -215,19 +216,25 @@ const Toast = ({
       </div>
     </div>
 
-    {onClick && buttonLabel &&
+    {!onClick && buttonLabel && url &&
```
thomasthesecond: For this whole block, we could do something similar to what we've done in Button (https://github.com/lonelyplanet/backpack-ui/blob/master/src/components/button/index.jsx#L197)

```
const Element = url ? "a" : "button";

{(url || onClick) && buttonLabel &&
  <Element href={url} onClick={onClick} style={[styles.button, styles.onClickButton]}>
    {buttonLabel}
  </Element>
}
```

Since `href` and `onClick` are `null` by default, the prop on the Element won't be applied if it is not set.

src/components/toast/index.jsx
```diff
@@ -148,116 +148,123 @@ const messageTypes = {
...
+            },
+          },
+        }}
+      />
 
-    {type && iconFromString(messageTypes[type].icon, {
-      style: styles.icon,
-      ariaHidden: true,
-    })}
+      {type && iconFromString(messageTypes[type].icon, {
```
nickfranciosi: @shawncothran looks like this block has too much indentation maybe?

shawncothran: It is lined up with the block above it, all good, view it raw

Issue #572, opened 2017-12-01
Inconsistent "Email me a link to sign in" button behavior


WunderBart:
Enabled & disabled states doesn't seem to be working well.

The button:
- enables after writing just one letter (at first attempt only),
- doesn't enable when proper email has been inserted via autofill,
- doesn't disable when email has been cleared (select all -> delete)

Example:
![lp sign in with email inconsistent button](https://user-images.githubusercontent.com/1451471/33485520-e7f1aee8-d6a6-11e7-9cf2-e4a11cb78536.gif)

Tested on Chrome.

laserchicken:
done

PR #577, opened 2017-12-05
fix: don't use valid and showErrors state

laserchicken:
request: https://github.com/lonelyplanet/backpack-ui/issues/572

### DIFF
--- a/src/components/modalContent/modalContentMagicLinkForm.jsx
+++ b/src/components/modalContent/modalContentMagicLinkForm.jsx
@@ -25,8 +25,6 @@ class MagicLinkForm extends Component {
     super(props);
     this.state = {
       value: "",
-      valid: false,
-      showErrors: false,
     };
 
     this.handleSubmit = this.handleSubmit.bind(this);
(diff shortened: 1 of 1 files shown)


laserchicken:
`handleChange` [here](https://github.com/lonelyplanet/backpack-ui/pull/577/files#diff-76c370f71ba5f247d46f462b02e69329R63) is called before `validate` so it always gets `errorCount` from previous validation and re-renders component with incorrect error state. You can see it in the example @WunderBart attached to the issue:
33485520-e7f1aee8-d6a6-11e7-9cf2-e4a11cb78536
You type "b", it calls `handleChange` with initial `errorCount = 0`, sets `state.valid=true` and re-renders - submit button is enabled. Then it calls `validate`, `errorsCount=1`, because `b` is incorrect email address. `validate` forces re-render but `state.valid` is `false`, so submit button stays enabled. 
Then you select correct email from the dropdown, it calls `handleChange` with previous `errorCount=1` - re-renders and submit button is disabled. `validate` is called, sets `errorCount=0` - it is correct address. 
You click outside input, `onBlur` is called and re-renders with previous `errorCount=0`, submit button is enabled again.

The problem here is colliding error states. I removed `state.valid`, `Validate` component should be single source of truth here.

PR #583, opened 2017-12-12
update: allow newsletter form to auto optin with new legal requirements

nickfranciosi:
- [x] updates default legal opt in text
- [x] allows opt in to be optional for default opt in

### DIFF
--- a/src/components/newsletter/index.jsx
+++ b/src/components/newsletter/index.jsx
@@ -39,12 +39,12 @@ const styles = {
   heading: {
     fontSize: "20px",
     letterSpacing: "-.3px",
-    lineHeight: (32 / 20),
+    lineHeight: 32 / 20,
 
     [`@media (min-width: ${media.min["480"]})`]: {
       fontSize: "24px",
       letterSpacing: "-.4px",
-      lineHeight: (32 / 28),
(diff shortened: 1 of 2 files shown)


src/components/newsletter/index.jsx
```diff
@@ -193,15 +193,15 @@ class Newsletter extends Component {
...
+    const { endpoint, signup, hasOptin } = this.props;
 
     this.setState({ waiting: true });
 
     const data = {
       newsletter: {
         [signup.vars.replace(/newsletter\[(.*)\]/, "$1")]: true,
         source: signup.source,
-        legalOptIn: this.state.acceptLegalOptIn,
+        legalOptIn: hasOptin ? this.state.acceptLegalOptIn : "legalOptIn",
```
laserchicken: So it sends "legalOptIn" string as `legalOptIn` value in case `hasOptin` is `false/null`? It looks inconsistent to me, we should either send boolean value or omit this field (do not include in the payload). What do you think?

nickfranciosi: I am just replicating what has existed and working - I think it has to be a string regardless. I can change it to `"true"` though. 

If you look at the value of the checkbox it is set to `"legalOptIn"`

PR #586, opened 2017-12-20
Add enableWindowScrolling prop to ModalWrapper

thomasthesecond:
This prop gives the option to disable the `no-scroll` module and allow
the parent window to scroll.

Allowing the parent window to scroll fixes a bug when a captcha is
loaded in a modal window and the captcha puzzle is sometimes pushed
off-screen, especially on mobile-sized screens. Without the `no-scroll`
module loading, the captcha puzzle can properly position itself.


thomasthesecond:
image

PR #588, opened 2017-12-29
fix: replace settings.json imports with new styles

shawncothran:
Replacing all the settings.json imports with newer style js files in preparation to transition to css modules with a clean slate
It's a LOT of code to look over, mostly 1:1 replacement
Moved `components.header.hightMobile` inside `dimensions.js`


settings.json
```diff
@@ -1,185 +0,0 @@
-{
```
thomasthesecond: Let's keep settings.json for backwards compatibility. Removing it could break other apps that use it. It will likely be removed in the next major version as a breaking change.

src/components/ad/index.jsx
```diff
@@ -2,32 +2,34 @@ import React from "react";
...
+      backgroundColor: colors.textPrimary,
       marginLeft: "auto",
       marginRight: "auto",
       position: "relative",
       textAlign: "center",
     },
 
     framed: {
-      backgroundColor: "#f6f6f6",
+      backgroundColor: colors.borderPrimary,
```
thomasthesecond: Let's keep this value hardcoded for now. We use `#f6f6f6` in all of the ads and there's not yet a color variable for it yet.

src/components/ad/index.jsx
```diff
@@ -63,22 +65,22 @@ function Ad({ id, framed, className, style }) {
 
   if (framed) {
     return (
-      <div className="Ad-wrap clearfix">
+      <article className="Ad-wrap clearfix">
```
thomasthesecond: This can remain a `div`

src/components/takeover/index.jsx
```diff
@@ -144,25 +147,25 @@ class Takeover extends Component {
 
         header: {
           base: {
-            backgroundColor: color.white,
-            height: components.header.heightMobile,
+            backgroundColor: colors.bgPrimary,
+            height: "50px",
```
thomasthesecond: Could we move the component values to the dimensions.js file?

src/components/settingBlock/settingBlockHeader.jsx
```diff
@@ -2,21 +2,27 @@ import React from "react";
...
+    fontWeight: fontWeightMedium,
+    lineHeight: lineHeightReset,
     paddingTop: "16px",
   },
 
   subtitle: {
     color: colors.textSecondary,
-    fontSize: typography.fontSizeUppercase,
-    fontWeight: typography.fontWeightLight,
+    fontSize: fontSizeUppercase,
```
thomasthesecond: Need to do
```
fontSize: `${fontSizeUppercase}px`
```

src/components/sectionalNav/index.jsx
```diff
@@ -3,17 +3,22 @@ import PropTypes from "prop-types";
...
+import mq from "../../styles/mq";
+import spacing from "../../styles/spacing";
+import timing from "../../styles/timing";
+import zIndex from "../../styles/zIndex";
 import { rgb } from "../../utils/color";
+import font from "../../utils/font";
 
 const navigationSubHeight = "80px";
-const navigationSubHeightMobile = settings.components.header.heightMobile;
+const navigationSubHeightMobile = "50px";
```
thomasthesecond: Another instance where we could put this value inside of dimensions.js

PR #624, opened 2018-02-27
Updated IconButton use of radiums Style component to use classes instead of element ids

harterc1:
This broke things when multiple "React trees" were being rendered on a single page and both were using IconButton -- they'd reuse each others element ids and cause conflicts.

Also, removed some PropType strictness.

### DIFF
--- a/src/components/iconButton/index.jsx
+++ b/src/components/iconButton/index.jsx
@@ -190,7 +190,7 @@ IconButton.propTypes = {
   className: PropTypes.string,
   href: PropTypes.string,
   onClick: PropTypes.func,
-  size: PropTypes.number,
+  size: PropTypes.oneOf([32, 40, 56]),
   owns: PropTypes.string,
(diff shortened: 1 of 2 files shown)


src/components/iconButton/index.jsx
```diff
@@ -68,132 +66,131 @@ const fontSize = {
...
-    "Share",
-  ]).isRequired,
+  iconName: PropTypes.string.isRequired,
   label: PropTypes.string.isRequired,
   id: PropTypes.string,
   className: PropTypes.string,
   href: PropTypes.string,
   onClick: PropTypes.func,
-  size: PropTypes.oneOf([32, 40, 56]),
+  size: PropTypes.number,
```
thomasthesecond: I think I'd prefer the size here to remain `oneOf` so we don't introduce random one-off sizes. Is there a design that does not use one of these sizes, or is this causing a problem somewhere else?

PR #625, opened 2018-02-27
Update tag styles

jcreamer898:


src/components/tag/index.jsx
```diff
@@ -18,28 +18,31 @@ const hoverStyles = {
   },
 
   selected: {
-    backgroundColor: colors.textPrimary,
+    backgroundColor: "#F3F6F7",
```
lburdock: Should these colors be added to the colors style file?

PR #630, opened 2018-03-06
update tabbedNav to have no outline on the buttons

katelynp43:
Remove blue outline on tabbedNav button links


thomasthesecond:
Was this a request from designers? They're killing me with these requests to remove accessibility features from our UI.

katelynp43:
No, it wasn't from designers. I noticed it when I was working on the blog yesterday and prefer to have it removed. :)

thomasthesecond:
Then I'd recommend keeping it or using our gray dotted outline style found in utils/mixins.js. We'll be working on a better style for focus states soon hopefully.

katelynp43:
Hey @thomasthesecond, I think it would look best to remove it, since it doesn't act like a true button. When the outline is there, you loose the red underline that is on the active state. I don't think we need the accessibility outline or a gray dotted line like you previously suggested.  I am attaching a screenshot here so you can see what I am referring too. Do you have further thoughts after seeing this? screen shot 2018-03-08 at 1 10 07 pm

thomasthesecond:
The outline is useful for users who may use their keyboard to navigate the page. Unfortunately, browsers apply the focus state when clicked with a mouse; typically we don't want the outline for mouse users. However, it's useful for keyboard users so we keep it.

Here's a video where I navigate the buttons with my tab key and press enter to "click" on the button

https://cl.ly/323S2K3X1W3G

It's very hard to do this when then outline is removed because you don't know where you are or which button has focus.

If you don't like how the blue outline looks, I recommend using our gray dotted outline or changing the color of the text when focused.

thomasthesecond:
Also, not to :trollface: but http://www.outlinenone.com/ explains the importance of the outline.

katelynp43:
@thomasthesecond that is a great point of view! I appreciate you taking the time to explain this. I hadn't thought about it like that before. Thanks for sharing. I will delete this branch and PR and leave the outline as is. :)

PR #641, opened 2018-03-23
feat: Added VideoPlaylistWithSlider.onLoadVideo prop

harterc1:


src/components/video/videoPlaylistWithSlider.jsx
```diff
@@ -70,6 +70,11 @@ class VideoPlaylistWithSlider extends React.Component {
   }
 
   onLoadVideo = (video, autoplay) => {
+    const { onLoadVideo } = this.props;
+    if (onLoadVideo) {
+      onLoadVideo(video, autoplay);
```
nickfranciosi: my only comment would to not name the function passed in and the function within the class the same thing. I would rename the one within the class to something like `handleVideoLoad`

PR #652, opened 2018-04-05
feat: add global header shopping cart icon - ref: DOT-103

shawncothran:


shawncothran:
removes blocker for https://github.com/lonelyplanet/dotcom-web/pull/277 @thomasthesecond

PR #674, opened 2018-05-15
feat(flights-widget): add flight fetcher DOT-309

nickfranciosi:
- [x] pulls in HOC to fetch flight data
- [x] exports a `NoFlightResults` component


commitlint.config.js
```diff
@@ -1,11 +1,3 @@
 module.exports = {
   extends: ["@commitlint/config-conventional"],
-  rules: {
-    "subject-case": [2, "always", "sentence-case"],
-    "type-enum": [2, "always", [
-      "feat",
-      "fix",
-      "chore",
-    ]],
-  },
```
thomasthesecond: Why remove this?

nickfranciosi: because it wasn't letting me commit even with just "feat: add flight fetcher"

thomasthesecond: I had configured it to use sentence case, so `feat: Add flight fetcher`. I had always preferred sentence case, but if we want to agree on all lower case, then ok

PR #679, opened 2018-06-04
fix: add book card

laserchicken:


thomasthesecond:
Judging by your component folder structure, if this is a component only used for marketing/kids, is it possible to place the component in that repo? However, if it's going to be re-used outside of marketing/kids, I'd recommend placing the component alongside the others in the file structure.

laserchicken:
Good point. This component may be used in other marketing projects, not only in marketing app (copilot for example). I moved it to the components root directory. ^

PR #684, opened 2018-06-07
fix: add border switch (book card)

laserchicken:


src/components/cardBook/index.jsx
```diff
@@ -72,7 +73,7 @@ const CardBook = ({
     <div style={{ textAlign: "center" }}>
       <div style={textHeading6()}>{title}</div>
```
thomasthesecond: I know this wasn't part of your changes, but have you considered making this a heading element using the `Heading` component?

src/components/cardBook/index.jsx
```diff
@@ -72,7 +73,7 @@ const CardBook = ({
...
       <div style={[textHeading8(), { color: colors.textSecondary }]}>{subtitle}</div>
-      <div style={styles.border} />
+      {!hideBorder && <div style={styles.border} />}
```
thomasthesecond: This border `div` could maybe be an `hr`

laserchicken:
thanks @thomasthesecond updated ^

PR #698, opened 2018-08-16
[DOT-600] Select First Search Result if forceSelection Prop Enabled

johnwquarles:
[DOT-600](https://lonelyplanet.atlassian.net/browse/DOT-600) This code will:

- Alter the meaning of the `<Typeahead>` component's `forceSelection` prop.

To this point, when active, it had been checking for the following:
- Has the user already made a selection?
- Has the user entered text that exactly matches one of the (filtered) options?

And if neither one of those conditions were true (ie, the user had entered incomplete text and no selection has been made yet), it would cause the selection to revert to the provided `initialValue` prop.

In practice, this typically appears to be `null`.

Note that if one of the options _had_ been entered (exactly, case-sensitively), then the user's input _wasn't_ being reset to `initialValue`, but the option that they entered was never actually being selected.

Anyway, my thoughts behind repurposing this prop:

- The ticket's request to force selection of the first search result on blur fits in nicely with where this `forceSelection` logic was already occurring.
- And hey, we're forcing selection by inserting the first result. That's... exactly what that is.
- And we don't want to put in `initialValue` anymore, per the ticket, so
- Let's just repurpose the prop.
- We certainly still care if the user has already made a selection, so no change to that condition.
- But we don't care if the user has entered in an exact match for an existing option or not.
- If they _had_, great, we'll be inserting that exact option since it'll be the first in a list of one option. This is intuitive behavior.
- And if they _hadn't_ entered in an exact option, then there might be one or several options for whatever they _did_ enter. If so, choose the first.
- But if there are no search results, do nothing.

Getting this to work involves having `onBlur` change state, and then waiting until the state change is complete prior to invoking the parent's `onBlur` callback (if present).

I tried to make this happen with as few changes to what methods were returning as possible-- wound up making the component's internal `onBlur` handler async, whereas it wasn't before. This should be the only change (by definition it now returns a promise), and it should not matter (probably didn't need to be so conservative, really, but didn't want to wind up in situations where a promise _or_ a synchronous value could be returned by these internal methods, even if the return values were never used).

**Screenshots**

carrental

**Definition of Done**
  - [X] Reference and/or create git issue for the feature or bug related to this PR.
  - [ ] Ensure feature or fix is unit tested
  - [X] Jenkins runs branch tests
  - [ ] Update docs if needed
  - [ ] Make sure to get a 👍 up before merging
  - [ ] Get sign off from a designer if this is a new feature
  - [ ] Mobile and browser test this feature

**Notes**
- Use the squash and merge option when merging

**Todo**
Time permitting, I will add (either here or in another PR):
- [ ] Updates to storybook reflecting the changes here
- [ ] Tests reflecting the changes here

### DIFF
--- a/src/components/typeahead/typeahead.jsx
+++ b/src/components/typeahead/typeahead.jsx
@@ -15,10 +15,10 @@ import KeyEvent from "./utils/keyEvent";
  * Renders an text input that shows options nearby that you can use the
  * keyboard or mouse to select. Requires CSS for MASSIVE DAMAGE.
  */
+let count = 0;
+
 class Typeahead extends Component {
   static getInstanceCount() {
-    let count = 0;
-
     return ++count; // eslint-disable-line no-plusplus
   }
 
@@ -270,17 +270,37 @@ class Typeahead extends Component {
     return null;
   }
 
-  onBlur(event) {
+  async onBlur(event) {
     this.setState({
       isFocused: false,
     }, () => {
       this.onTextEntryUpdated();
     });
 
-    if (this.props.forceSelection &&
-        (this.state.searchResults.indexOf(this.state.entryValue) === -1) &&
-        this.state.selectionIndex === null) {
-      this.setState({ entryValue: this.props.initialValue });
+    const thereAreSearchResults = this.state.searchResults.length > 0;
+    const thereIsInput = event.target.value !== "";
+    /**
+     * If the user is not hovering or arrow-keying over any particular option, we may safely
+     * select the topmost option among the search results. It'll either be the one
+     * that they've already selected, or they otherwise are not disclosing any specific preference.
+     */
+    const noSpecificSelectionIsPending = [null, -1].includes(this.state.selectionIndex);
+    const insertFirstSearchResult = [
+      this.props.forceSelection,
+      noSpecificSelectionIsPending,
+      thereAreSearchResults,
+      thereIsInput,
+    ].every(Boolean);
+
+    if (insertFirstSearchResult) {
+      // Force selection of the first search result.
+      const [option] = this.state.searchResults;
+      const optionString = this.displayOption(option, 0);
+      await this.setSelectedOption(option);
+      if (this.props.validate) {
+        this.props.validate(optionString, this.state.searchResults);
+      }
+      this.props.onOptionSelected(event, option);
     }
 
     if (this.props.onBlur) {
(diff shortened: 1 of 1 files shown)


src/components/typeahead/typeahead.jsx
```diff
@@ -277,10 +277,12 @@ class Typeahead extends Component {
       this.onTextEntryUpdated();
     });
 
-    if (this.props.forceSelection &&
-        (this.state.searchResults.indexOf(this.state.entryValue) === -1) &&
-        this.state.selectionIndex === null) {
-      this.setState({ entryValue: this.props.initialValue });
+    const thereAreSearchResults = Boolean(this.state.searchResults.length);
```
egdelwonk: Fyi: `Boolean(1) === !!1 // true` 

Might not be as readable as `Boolean` though, but might be something you might come across.

johnwquarles: Ah, `this.state.searchResults.length > 0` is way more straightforward. Don't know why the other way occurred to me first. Will change.

johnwquarles:
I'm getting inconsistent results with this `backpack-ui` change when running `copilot` locally and bringing in the changes-- seems like clicking outside of the `<TypeAhead>` to blur always has the intended effect, but using the tab key does not (necessarily).

Will investigate on Monday.

thomasthesecond:
The changelog gets updated automatically when we make a new release, you don't have to manually update it.

johnwquarles:
@thomasthesecond heh, d'oh, thanks for letting me know!

src/components/typeahead/typeahead.jsx
```diff
@@ -291,37 +311,13 @@ class Typeahead extends Component {
   }
 
   onOptionSelected(event, option) {
+    const optionString = this.displayOption(option, 0);
```
vacas5: looks like this is happening in the `setSelectedOption` as well. Does it need to be?

src/components/typeahead/typeahead.jsx
```diff
@@ -270,17 +270,37 @@ class Typeahead extends Component {
...
+      this.props.forceSelection,
+      noSpecificSelectionIsPending,
+      thereAreSearchResults,
+      thereIsInput,
+    ].every(Boolean);
+
+    if (insertFirstSearchResult) {
+      // Force selection of the first search result.
+      const option = this.state.searchResults[0];
+      const optionString = this.displayOption(option, 0);
```
vacas5: repeated inside the setSelectedOption

src/components/typeahead/typeahead.jsx
```diff
@@ -270,17 +270,37 @@ class Typeahead extends Component {
...
+    const insertFirstSearchResult = [
+      this.props.forceSelection,
+      noSpecificSelectionIsPending,
+      thereAreSearchResults,
+      thereIsInput,
+    ].every(Boolean);
+
+    if (insertFirstSearchResult) {
+      // Force selection of the first search result.
+      const option = this.state.searchResults[0];
```
egdelwonk: We might prefer to use destructuring here instead of access the array by index. Something like 
`const [option] = this.state.searchResults;` instead.

src/components/typeahead/typeahead.jsx
johnwquarles: Ah, my initial concern was unfounded-- it's using the return value of a getter, which is a function, and passing arguments to it.

The reason that `displayOption` is called twice is because its return value (`optionString`) is needed in both places-- `setSelectedOption` does a number of things with it, and `onOptionSelected` uses it to pass to an optional validation callback from the parent.

I didn't want to have to pass `optionString` to `setSelectedOption` in addition to `option`, and didn't think that it was worth breaking out `this.displayOption(option, 0)` into its own function or method, since it more or less already is just that.

What `displayOption` actually _is_ (beyond simply "a function") can vary, so I'm not sure that the second argument of "0" is ever used, but wanted to err on the side of caution and leave it alone.

PR #699, opened 2018-08-27
refactor: DateRange component outside range test

vacas5:
This began as a bug fix but I think I determined that it was in
actuality an issue with the DateRangePicker dependency and not our
code, after implementing tests for our code. Adding tests allowed me to
refactor in a way that seems a bit more straightforward. Also fixed an
issue with the component that was firing React error messages in the
console.

Introduced enzyme as a library to help test React components, and
refactored existing tests a bit for improved visualization.

### DIFF
--- a/spec/components/form/daterange.spec.js
+++ b/spec/components/form/daterange.spec.js
@@ -1,0 +1,116 @@
+import React from "react";
+import moment from "moment";
+import { expect } from "chai";
+import { configure, shallow } from "enzyme";
+import { DateRange } from "../../../src";
+import Adapter from "enzyme-adapter-react-16.2";
+
+configure({ adapter: new Adapter() });
+
+describe("DateRange component", () => {
+  it("renders without crashing", () => {
+    shallow(<DateRange />);
+  });
+
+  describe("isOutsideRange determination", () => {
+    it("returns true if date argument is before today", () => {
+      const component = shallow(<DateRange />);
+      const date = moment("2018-08-15");
+      const actual = component.instance().isOutsideRange(date);
+      expect(actual).to.equal(true);
+    });
+
+    it("returns true if date in the future but is past the lastSelectable date", () => {
+      const lastSelectableDate = moment().add(7, "days");
+      const component = shallow(
+        <DateRange lastSelectableDate={lastSelectableDate} />
+      );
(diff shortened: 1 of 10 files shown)


spec/utils/mixins.spec.js
```diff
@@ -11,164 +11,166 @@ const baseStyles = {
...
-    expect(styles).to.deep.equal({
-      border: "1px solid red",
-      padding: "10px",
-      width: "auto",
-      outline: "1px lightgray dotted",
-      outlineOffset: "2px",
+describe('mixins', () => {
+  describe("outline mixin", () => {
+    it("should add outline styles with a default outline offset of 2", () => {
+      const styles = Object.assign({}, baseStyles, outline());
```
johnwquarles: Just as a thought, we could introduce rest spread here.

spec/components/form/daterange.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  describe("isOutsideRange determination", () => {
+    it("returns true if date argument is before today", () => {
+      const component = shallow(<DateRange />);
+      const date = moment("2018-08-15");
+      const actual = component.instance().isOutsideRange(date);
+      expect(actual).to.equal(true);
+    });
+    it("returns true if date in the future but is past the lastSelectable date", () => {
```
johnwquarles: Dunno if it's a convention per se, but other tests I've seen are spacing between each individual test (`it`) but not immediately after a group (`describe`).

spec/utils/mixins.spec.js
vacas5: basically just wrapped a whole lot of existing code in an additional describe block for the test console output. This isn't new code. We can/should refactor later

spec/components/form/daterange.spec.js
vacas5: I formatted these tests using prettier, and it collapsed them all. So we may need to adjust the prettier settings but we should decide on a whole for all of our repos

spec/utils/mixins.spec.js
vacas5: accidentally deleted my follow up comment: this was simply me wrapping all of these tests in another describe block to make the console output more readable. There's probably a lot of room for refactoring these down the road

PR #701, opened 2018-08-31
fix: truncate button text overflow

johnwquarles:
### This code will:

Truncate button text overflow.

The PR was originally also going to remove left/right padding from a full-width button, but @vacas5 and I concluded that it'd be best (until we know more about the design intent behind that padding) to leave it alone.

I will make another PR to specifically address the issue occurring in [DOT-499](https://lonelyplanet.atlassian.net/browse/DOT-499) where it occurs.

Adding the text truncation here _does_ address an existing issue, though, so figure it's worth leaving in place. I'll add a ticket for it shortly.

**Definition of Done**
  - [X] Reference and/or create git issue for the feature or bug related to this PR.
  - [ ] Ensure feature or fix is unit tested
  - [ ] Jenkins runs branch tests
  - [ ] Update docs if needed
  - [ ] Make sure to get a 👍 up before merging
  - [ ] Get sign off from a designer if this is a new feature
  - [ ] Mobile and browser test this feature


johnwquarles:
Seems to overflow. Both the button text & background are white, so you have to change the background color to see it, but it does.

I will add `overflow: hidden` and `text-overflow: ellipsis` to deal with it.

johnwquarles:
Updated this to address overflow solely by means of truncating the text, so as to avoid possibly running afoul of design intent.

PR #704, opened 2018-11-01
feat: track newsletter subscriptions per AP spec

johnwquarles:
Imports `lp-analytics` and uses it to push a very simple event to the dataLayer whenever a newsletter subscription occurs.

Note that inclusion of `gtm.js` (and the actual tracking activity that its inclusion establishes) is up to the host codebase; this will only ensure that events are pushed to the `dataLayer` (which will be created if necessary). In the absence of `gtm.js` it's a plain ol' JS array. In the presence of `gtm.js` it's a patched array whose `.push` method also sends events along to gtm.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -393,6 +393,28 @@
         }
       }
     },
+    "@lonelyplanet/lp-analytics": {
+      "version": "2.0.6",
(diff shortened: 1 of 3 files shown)


johnwquarles:
A question for whoever may know: will this cover all newsletter subscriptions? Do legacy versions of this widget exist, or is the functionality perhaps implemented by other means anywhere else?

PR #710, opened 2018-12-18
fix: lock version of react-measure

johnwquarles:
I think this is breaking `destinations-next` builds.

Jenkins keeps throwing `Module not found: Error: Can't resolve '@babel/runtime/helpers/esm/objectWithoutPropertiesLoose' in '/mnt/workspace/destinations-next-branch-test/node_modules/react-measure/dist'`.

Seems like the same issue [cropped up](https://github.com/meteor/meteor/issues/10128#issuecomment-410523377) in several big-name JS projects this August, when Babel 7 was released.

 And `react-measure` [upgraded to Babel 7](https://github.com/souporserious/react-measure/commit/14e725e496fb0ac34f0ada7b011d134d4ee0872d#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L50) in a commit that was released with its version 2.2.

[`react-measure`](https://www.npmjs.com/package/react-measure/v/2.2.0) 2.2 was published 4 days ago at time of writing, which aligns with the timing of when Jenkins started breaking in `destinations-next`.

So,

### the solution:

lock `react-measure` to `2.1.3`, at least for now.

I suppose a better solution may be to update node & the Jenkins configuration in `destinations-next` to respect `package-lock.json`, but this is a quicker fix, and I wouldn't be surprised if `destinations-next` weren't the only victim of this issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11988,13 +11988,13 @@
       }
     },
     "react-measure": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/react-measure/-/react-measure-2.0.2.tgz",
-      "integrity": "sha1-ByqaX6/AHfutwfpfsJ/DUQN/Y2w=",
+      "version": "2.1.3",
+      "resolved": "https://registry.npmjs.org/react-measure/-/react-measure-2.1.3.tgz",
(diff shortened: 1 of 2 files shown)


johnwquarles:
(it worked!)

PR #714, opened 2019-01-04
fix: add optional prop for image alt text

katelynp43:
Added optional prop for image alt text to the ArticlePreview component. This will help aid in accessibility of images.


katelynp43:
@thomasthesecond Yes, of course! Just added a commit with default props.

PR #722, opened 2019-01-08
Use button for legacy log in link in MultiStepLogin

thomasthesecond:
The linter throws an error when using an anchor without a proper `href`
value.

```
Links must not point to "#". Use a more descriptive href or use a button
instead  jsx-a11y/href-no-hash
```

This error prevents a release from being made.

Related #720

### DIFF
--- a/src/components/multiStep/multiStepLogin.jsx
+++ b/src/components/multiStep/multiStepLogin.jsx
@@ -27,6 +27,10 @@ const styles = {
     color: colors.textPrimary,
     fontWeight: fontWeightMedium,
   },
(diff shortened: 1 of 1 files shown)


nidalaa:
Hi @thomasthesecond unfortunately this button looks pretty bad:

screen shot 2019-01-09 at 14 45 43

The question is:
- should we add background-color: white to its CSS styles?
- or try `more descriptive href` and go back to a element?

thomasthesecond:
My fault I missed this! Background color is the best solution.

PR #723, opened 2019-01-09
fix: change background for username+password login link

nidalaa:


nidalaa:
@thomasthesecond  is in comment here: https://github.com/lonelyplanet/backpack-ui/pull/722#issuecomment-452701911

src/components/multiStep/multiStepLogin.jsx
```diff
@@ -28,6 +28,7 @@ const styles = {
     fontWeight: fontWeightMedium,
   },
   legacyButton: {
+    "background-color": "white",
```
thomasthesecond: Can we do `backgroundColor: colors.bgPrimary` here?

nidalaa: 👍

nidalaa:
Fixed, thanks! :)

PR #731, opened 2019-01-24
fix: lineclamp is expecting a number but is given a bool

Courey:
lineclamp is being passed false but is expecting a number
 https://github.com/lonelyplanet/backpack-ui/blob/8dc8c7e430f3f68e55edc7e3f18c7602547c60fc/src/components/thumbnailListItem/index.jsx#L309

### DIFF
--- a/src/components/video/videoPlaylistWithSlider.jsx
+++ b/src/components/video/videoPlaylistWithSlider.jsx
@@ -201,7 +201,7 @@ class VideoPlaylistWithSlider extends React.Component {
                           onClick={!followVideoUrls ? () => this.onClickVideo(v) : null}
                           imagePath={v.thumbnailImage}
                           subtitle={[duration(v.duration)]}
-                          lineClamp={false}
+                          lineClamp={0}
                         />
                       ))}
                     </ThumbnailList>


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "backpack-ui",
-  "version": "4.90.17",
+  "version": "4.90.18",
```
vacas5: so, there's actually a more formal release process for this package. We'll run some npm commands and this file will be updated automagically based on the types of commits that were made to the repo. So, you should actually *not*  change this file

src/components/video/videoPlaylistWithSlider.jsx
```diff
@@ -201,7 +201,7 @@ class VideoPlaylistWithSlider extends React.Component {
                           onClick={!followVideoUrls ? () => this.onClickVideo(v) : null}
                           imagePath={v.thumbnailImage}
                           subtitle={[duration(v.duration)]}
-                          lineClamp={false}
+                          lineClamp={0}
```
vacas5: must be a regression introduced when ThumbnailListItem was enhanced. My God we could use tests. 👍

package.json
Courey: ok, sorry. I'll snag that out then.

PR #736, opened 2019-02-01
fix: add analytics click trackers to article preview component

katelynp43:
Adding click event tracker for the article preview component. By adding this, in each repo that this component is being used, we can add a name and id to the event tracker so that these are viewable in GA. Per Kristen and Analytics Pros, we should use "Promotion Click" for this type of even tracking.

@johnwquarles can you please confirm that this is working as it should? I was using that chrome extension but it appears that it doesn't work on localhost, or perhaps I am missing something. Can you please let me know the best way to test this?

### DIFF
--- a/src/components/articlePreview/index.jsx
+++ b/src/components/articlePreview/index.jsx
@@ -2,6 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import assign from "object-assign";
 
+import { analytics, EventNames, AnalyticsEvent } from "@lonelyplanet/lp-analytics";
 import colors from "../../styles/colors";
 import font from "../../utils/font";
 import CategoryLabelLink from "../categoryLabelLink";
@@ -9,7 +10,17 @@ import Heading from "../heading";
 
 const markup = html => ({ __html: html });
 
-function ArticlePreview({ title, paragraph, image, imageAlt, href, category, categoryHref }) {
+function ArticlePreview({
+  title,
+  paragraph,
+  image,
+  imageAlt,
+  href,
+  category,
+  categoryHref,
+  trackEventName,
+  trackEventPosition,
+}) {
   const styles = {
     container: {
       fontFamily: font("benton"),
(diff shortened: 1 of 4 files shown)


johnwquarles:
@katelynp43 

Sure, I just got done taking a look.

Here are my notes:

- I think that the difficulty in confirming that the events were firing was due to `Storybook`'s use of iframes. The browser extension probably (and reasonably) checks `window.dataLayer` directly.

I have found `window.frames[0].dataLayer` to work. When checking there, I see that the events are being populated as expected 👍 (added `e.preventDefault` to give myself enough time to check)

- Couldn't hurt to add those properties in `stories/index.jsx` just so that they are editable widgets-- I see that default values are provided, so I know that it'll work regardless, but hey, it's just two lines

```
      trackEventName={text("Tracking event name")}
      trackEventPosition={text("Tracking event position")}
```

- I remember thinking that `promotionClick` has a weirdly specific shape.

Given that, I think it'd be great to add a `createPromotionClick` method in `lp-analytics` that accepts the only two dynamic values and spits out the fully-formed object.

Could just add it into `src/events/index.ts`.

I think this PR is fine as-is, but if you'd like to use that `createPromotionClick` you can let me know and I'll make & publish it and you can add it here before merging this PR. Or feel free to take it on yourself.

katelynp43:
@johnwquarles I think it would be great to add `createPromotionClick`. Let me know when you get it added to lp-analytics and released, then I will update my PR here :) 

I believe there is an id pair available as well. I was on the fence about how a naming convention would work for the name and id, but I think we need to include it. I will update the PR to include all four of these parameters. ;)

'id': '<promotion id>',
'name': '<promotion name>',
'creative': '<promotion creative>',
'position': '<promotion position or page>'

katelynp43:
@johnwquarles I was able to swap out the event and confirm. Thanks for adding it to lp-analtyics.

screen shot 2019-02-05 at 12 01 29 pm

src/components/articlePreview/index.jsx
```diff
@@ -52,30 +64,60 @@ function ArticlePreview({ title, paragraph, image, imageAlt, href, category, cat
...
+            <CategoryLabelLink href={categoryHref}>{category}</CategoryLabelLink>
+            <a
+              href={href}
+              style={assign({}, styles.anchor, { marginTop: "12px" })}
+              onClick={() => {
+                track(
+                  createPromotionClickEvent({
+                    id: trackEventId,
+                    name: trackEventName,
+                    creative: "article preview text",
```
johnwquarles: Just for context (and forgive me if you're already aware)-- the `promotionClick` and `promotionImpression` events were originally meant to track ads from DFP (aka DoubleClick aka Google Ad Manager), which have a unit called a "creative".

Ultimately the decision was made that DFP's internal tracking was sufficient, and these events were n't used until now.

I think it's a good call to repurpose the key/value pair like this.

PR #739, opened 2019-02-12
fix: console errors, updated validate dep

vacas5:
Fixed several console issues related to using the input and Validate
component, by cleaning up which props could be spread on the input.
Unfortunately this was spreading dom event handlers like onBlur also,
and that's expected behavior even though not defined in the Input api.

Additionally, updated the version of the react-validate-form so we no
longer get dependency errors for that package.


src/components/form/input.jsx
```diff
@@ -5,22 +5,21 @@ import styles from "./styles";
...
-    error,
-    size,
-    theme,
-    fill,
-    customStyles,
-  } = props;
+  const { name, id, type, error, size, theme, fill, customStyles } = props;
+
+  // remove custom props that throw console errors
+  const sanitizedProps = Object.keys(props).reduce((iter, key) => {
```
johnwquarles: Just to note, this seems like a good use case for `Object.entries`

```javascript
Object.entries(obj).reduce((acc, [key, value]) => { ...code goes here }, {});
```

or something in `lodash`.

vacas5: TIL. Object.entries is perfect.

PR #751, opened 2019-04-09
fix: add additional qa hook functionality

dmhalejr:
`qaHook` is not valid (and ignored) on `react-router` Link objects and caused a warning in the console for dotcom-frontend. That has been fixed to handle via destructuring the `props` object being passed into our `Link` object that has a ternary that appends the external link `data-qa` hook properly or straight up ignores it. 

Additional QA hooks have been added to enhance testability in dotcom-frontend. All made optional inside backpack.

### DIFF
--- a/src/components/link/index.jsx
+++ b/src/components/link/index.jsx
@@ -1,22 +1,22 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Link as RouterLink } from "react-router";
+import createQAHook from "../../utils/createQAHook";
 
 const isExternal = (url) => /^(http|https):\/\//.test(url || "");
 
-const Link = (props) => (
-  isExternal(props.to) || (!props.to && props.onClick) ?
+const Link = ({ to, onClick, children, qaHook }) => (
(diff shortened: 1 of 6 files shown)


src/components/segment/index.jsx
```diff
@@ -85,6 +87,11 @@ Segment.propTypes = {
     PropTypes.element,
     PropTypes.arrayOf(PropTypes.element),
   ]),
+  qaHook: PropTypes.string,
+};
+
+Segment.defaultProps = {
+  qaHook: null,
```
vacas5: did we default the other string types to `""`?

src/components/link/index.jsx
```diff
@@ -28,11 +28,11 @@ Link.propTypes = {
     PropTypes.node,
     PropTypes.element,
   ]),
-  qaHook: PropTypes.bool,
+  qaHook: PropTypes.string,
 };
 
 Link.defaultProps = {
-  qaHook: false,
+  qaHook: null,
```
vacas5: see other comment about string default

PR #762, opened 2019-05-03
fix: PhotoGallery CSS updates

harterc1:
**1.** z-index increased to `2000` (same z-index as our site header in `dotcom-frontend`) so that our header doesn't appear over the photo gallery overlay.  react-photoswipes default z-index is `1500`.  This isn't really a noticeable problem because the clickable photos on our news gallery pages are far enough down the page that you never see that the site header is on top of the photo gallery.

**2.** Fixed captions having a semi-transparent black background on mobile.
RPReplay_Final1557170130 2019-05-06 15_40_04

**3.** PhotoGallery navigation arrows and the close button in the upper right weren't appearing.

They were most likely working at one point before the majority of browsers updated themselves to more strict standards:

Two changes:
1. For IE support: Needed to replace `;utf8,` with `;charset=utf8,`.
2. The SVG path needs to be URIEncoded.

Reference:
https://stackoverflow.com/questions/32870788/css-using-raw-svg-in-the-url-parameter-of-a-background-image-in-ie

Preview of arrows and 'x' button:
Screen Shot 2019-05-03 at 5 08 07 PM

### DIFF
--- a/src/components/photoGallery/styles.js
+++ b/src/components/photoGallery/styles.js
@@ -16,7 +16,7 @@ export default `
 
   .pswp__button--close {
     background-color: transparent;
-    background-image: url('data:image/svg+xml;charset=utf8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%2232px%22%20height%3D%2232px%22%20x%3D%220px%22%20y%3D%220px%22%20viewBox%3D%220%200%2032%2032%22%3E%3Cpath%20fill%3D%22%23d3d3d3%22%20d%3D%22M17.6%2C16L31.7%2C1.9c0.4-0.4%2C0.4-1.2%2C0-1.6c-0.4-0.4-1.1-0.4-1.6%2C0L16%2C14.4L1.9%2C0.3c-0.4-0.4-1.2-0.4-1.6%2C0c-0.4%2C0.4-0.4%2C1.2%2C0%2C1.6L14.4%2C16L0.3%2C30.1c-0.4%2C0.4-0.4%2C1.2%2C0%2C1.6c0.4%2C0.4%2C1.2%2C0.4%2C1.6%2C0L16%2C17.6l14.1%2C14.1c0.4%2C0.4%2C1.2%2C0.4%2C1.6%2C0c0.4-0.4%2C0.4-1.1%2C0-1.6L17.6%2C16z%22%20%2F%3E%3C%2Fsvg%3E') !important;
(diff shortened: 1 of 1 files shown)


src/components/photoGallery/styles.js
```diff
@@ -63,12 +67,21 @@ export default `
 
   .pswp__button--arrow--left:before {
     left: 0;
-    background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="21px" height="36px" x="0px" y="0px" viewBox="0 0 21 36"><path fill="#d3d3d3" d="M20.6,33.6L3.6,18l17-15.6c0.6-0.6,0.6-1.5,0-2c-0.6-0.6-1.5-0.6-2,0L0.6,16.8c-0.1,0-0.2,0-0.2,0.1C0.1,17.2,0,17.6,0,18c0,0.4,0.1,0.8,0.4,1.1c0.1,0.1,0.2,0,0.2,0.1l17.9,16.4c0.6,0.6,1.5,0.6,2,0C21.1,35,21.1,34.1,20.6,33.6z" /></svg>') !important;
+    background-image: url('data:image/svg+xml;charset=utf8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%2221px%22%20height%3D%2236px%22%20x%3D%220px%22%20y%3D%220px%22%20viewBox%3D%220%200%2021%2036%22%3E%3Cpath%20fill%3D%22%23d3d3d3%22%20d%3D%22M20.6%2C33.6L3.6%2C18l17-15.6c0.6-0.6%2C0.6-1.5%2C0-2c-0.6-0.6-1.5-0.6-2%2C0L0.6%2C16.8c-0.1%2C0-0.2%2C0-0.2%2C0.1C0.1%2C17.2%2C0%2C17.6%2C0%2C18c0%2C0.4%2C0.1%2C0.8%2C0.4%2C1.1c0.1%2C0.1%2C0.2%2C0%2C0.2%2C0.1l17.9%2C16.4c0.6%2C0.6%2C1.5%2C0.6%2C2%2C0C21.1%2C35%2C21.1%2C34.1%2C20.6%2C33.6z%22%20%2F%3E%3C%2Fsvg%3E') !important;
```
johnwquarles: I followed a few links from the StackOverflow reference you posted-- one point brought up [here](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris) is that you can save some space in the urlEncoded version by wrapping in `url(" ")` (double-quotes) and using single-quotes for the svg's attributes within it, since single quotes don't change when encoding. ie, every "`%22`" just becomes "`'`".

src/components/photoGallery/styles.js
```diff
@@ -97,13 +110,4 @@ export default `
     color: ${rgba(colors.textPrimary, 0.8)};
     text-decoration: underline;
   }
-
-  .pswp__top-bar {
-    height: 0;
-  }
-
-  .pswp__ui--fit .pswp__top-bar,
-  .pswp__ui--fit .pswp__caption {
```
johnwquarles: Was `!important` needed because these selectors lost specificity when dropping `.pswp__ui--fit` ?

harterc1: `pswp__ui--fit` doesn't always get applied to the parent container (seems to depend on the screen size + the mobile device you're on 🤷‍♂ ).  I removed use of `!important` in favor of just more specific selectors.

PR #768, opened 2019-06-07
fix: add babel-loader

laserchicken:
1. Recently updated storybook/react needs babel-loader https://github.com/storybookjs/storybook/issues/4116

2. Remarkable (dependency of react-markdown-renderer) has security vulnerability, I guess it's not gonna get patched anytime soon (see. https://github.com/jonschlinkert/remarkable/issues/331). We can replace it with `markdown-react-js`, I tested in storybook and it looks ok.


vacas5:
@laserchicken do we know which part of the site uses the component that uses the markdown rendering? somewhere in profile I think, so maybe dotcom-profile. Could we try running locally just to make sure there aren't any visual regressions?

laserchicken:
@vacas5 It's used on public profile pages, intro field [here](https://github.com/lonelyplanet/dotcom-profile/blob/7aea7936b4ce4c2d134de4f59e62ebe599e0d9cd/app/profile/containers/publicProfile.jsx#L197). Tested locally in dotcom-profile, it looks good:
Screen Shot 2019-06-10 at 4 27 31 PM

PR #782, opened 2020-01-21
feat: enable intrepid and ua activities

laserchicken:
https://lonelyplanet.atlassian.net/browse/CON-153


laserchicken:
thanks @vacas5

PR #783, opened 2020-03-25
fix: text area onFocus

laserchicken:
see https://lonelyplanet.slack.com/archives/C026V0UT3/p1585150833000400
Hannah Horn reported:
"It seems that there might be a bug which is stopping users from being able to update their LP profile bio/intro - @john falkum and I tested this and we were also both unable to edit the Intro section. I can click in the box but when I try to type anything, it doesn't work :confused:. I am still able to edit and save a user's bio from within the auth/luna part of a user's profile."

### DIFF
--- a/src/components/form/textarea.jsx
+++ b/src/components/form/textarea.jsx
@@ -15,6 +15,8 @@ function TextArea(props) {
     theme,
     customStyles,
     qaHook,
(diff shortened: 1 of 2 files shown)


laserchicken:
I cant' merge it @vacas5

PR #786, opened 2021-04-15
fix: props not being passed to Link component when using relative url

harterc1:
Problem was introduced here: 
https://github.com/lonelyplanet/backpack-ui/commit/ceada46f075787933343799f74902e80d233c25a#diff-9e7d4605239820b84fbe18a8dc471ad28e4c829b4ef712f22ce25802dfc638e3R18

React-router doc saying it's `Link` component handles standard props just fine: https://reactrouter.com/web/api/Link/others

The props for the `Link` component were using a spread to pass any remaining props into the underlying `a` tag, but only for internal links.

This created a problem because some components rely on styling the underlying `a` tag.  This was breaking padding and absolute positioning of children. 

Before (notice horizontal padding issue between image and text -- the image is wrapped in a `Link` component ):
Screen Shot 2021-04-15 at 12 09 59 PM

After:
Screen Shot 2021-04-15 at 12 56 55 PM

### DIFF
--- a/src/components/link/index.jsx
+++ b/src/components/link/index.jsx
@@ -6,7 +6,7 @@ import { validReactAttributes } from "../../utils/validReactAttributes";
 
 const isExternal = (url) => /^(http|https):\/\//.test(url || "");
 
-const Link = ({ to, onClick, children, qaHook, className, ...rest }) => {
+const Link = ({ to, onClick, children, qaHook, ...rest }) => {
   const sanitizedProps = validReactAttributes(rest);
 
   return (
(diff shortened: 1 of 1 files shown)


vacas5 (approved):
code looks fine, but does this solve that invisible div issue as well?

harterc1:
> code looks fine, but does this solve that invisible div issue as well?

Yes, it's because the outer `a` tag is supposed to have `position: relative` on it, but the styling isn't being applied because the `style` prop isn't passed down.    So, the child `div` has `position: absolute` on it and it's covering everything on the page until it gets to a parent div that has `position: relative` on it.

